PR #13, opened 2018-04-01
Blazor components topic

### LINKED ISSUE #8: Blazor components topic: Index topic
Title: Blazor components
UID: client-side/blazor/components/index
URL: https://docs.microsoft.com/aspnet/core/client-side/blazor/components

* Using Razor
* Inline vs code behind
* Event handling
* Data binding and validation
* Rendering child content
* Providing parameters
* Specifying child content
* Routing
* JS/TS interop

guardrex:
Fixes #8 

* 🐕 *ROUGH! ROUGH!* 🐕 draft for sure. I wrote to our outline, but I'm sure I'm missing bits. I made educated guesses in spots ... and they really weren't very "educated" given only a couple of days thus far with the tech.
* Follows our plan from the call regarding Razor ... I show what works. I mention our Razor topic, but I don't tell the reader go learn everything in the Razor topic.
* To show Blazor, C#, Razor, and HTML, I start with a carefully documented example, and then I break it down section-by-section for each piece. That example at the top of the topic is also in our ...
* :tada: **_Components Sample app!_** 🎈, which demos many of the features described in the topic. I probably need to link another line at the top back out to our .NET site *Get started* page with setup instructions. [btw- I still have a concern about not having a *Get started* page in the doc set.]
* This topic is set up as an index page in a "Blazor components" TOC area: In the future, many of the sections in this topic become expanded into *full topics* under this node of the TOC. The sections here will then link to those new topics for further (detailed) content.


docs/components/index.md
```diff
@@ -0,0 +1,310 @@
...
+## JavaScript/TypeScript interop
+
+To call JavaScript libraries or custom JavaScript/TypeScript code from .NET, the current approach is to register a named function in a JavaScript/TypeScript file:
+
+```javascript
+Blazor.registerFunction('doPrompt', message => {
+  return prompt(message);
+});
+```
+
```
muqeet-khan: @guardrex  may be consider mentioning that the JS file or <script> tag with registerFunction invocations should be placed in index.html after the blazor boot script tag and not from the components\Razor themselves. Else you risk running into issues like https://github.com/aspnet/Blazor/issues/451

guardrex: yes, ty @muqeet-khan I'll do that. :+1: I'll get in there on the next commit.

docs/components/common/samples/2.x/ComponentsSample/Pages/Index.cshtml
```diff
@@ -0,0 +1,84 @@
...
+    public string HeadingFontStyle { get; set; } = "normal";
+    public bool ItalicsCheck { get; set; } = false;
+
+    // When UpdateHeading is executed:
+    //    HeadingText is assigned the value of NewHeadingText.
+    //    ItalicsCheck determines the value of HeadingFontStyle to 
+    //        set the font style of the heading.
+    //    WebUtility.HtmlEncode ensures that user input can't 
+    //        compromise the security of the page with a malicious 
+    //        entry.
```
SteveSandersonMS: The HtmlEncode stuff mentioned here is misleading. It's not necessary. It won't avert XSS attacks, because XSS attacks aren't applicable here in the first place, plus having it would lead to content being double-encoded in the UI.

Could you remove it? Thanks!

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+By [Luke Latham](https://github.com/guardrex)
+
+[!INCLUDE[](~/includes/blazor-preview-notice.md)]
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/client-side/blazor/components/common/samples/)
+
+To download the sample:
+
+1. Download and unzip the [Blazor.Docs repository zip file](https://codeload.github.com/aspnet/Blazor.Docs/zip/master).
+1. Use the URL in the [sample link](https://github.com/aspnet/Docs/tree/master/aspnetcore/client-side/blazor/components/common/samples/) to navigate to the sample folder.
```
SteveSandersonMS: I might be being dim, but I don't understand this second list item at all. What URL? What sample folder? And how do you *run* the samples after you've downloaded them?

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+A Blazor *component* represents any feature or collection of features that can be represented by a [Razor file (\*.cshtml)](https://docs.microsoft.com/aspnet/core/mvc/views/razor) or a [C# file (\*.cs)](/dotnet/csharp/getting-started/) compiled into a C# class assembly. Components are typical features found in web app client UIs, such as a page, dialog, or form, along with its processing logic. A component can also consist entirely of programming logic without UI elements. Components are the fundamental building blocks of Blazor apps and can be nested, reused, and shared between projects.
```
SteveSandersonMS: The first sentence is quite complex. How about:

Blazor applications are built using *components*. A Blazor component is a self-contained chunk of user interface (UI), such as a page, dialog, or form. It includes both the HTML markup to be rendered along with processing logic needed to inject data or respond to UI events. Typically, you'll write components using Razor files (`*.cshtml`). Components are flexible and lightweight, and can be nested, reused, and shared between projects.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+When the index page loads, the `HeadingComponent` loads. The component:
+
+* Displays a heading on the page.
+* Presents a form that allows the user to change the heading text and control its font style.
+
+Heading component (*HeadingComponent.cshtml*):
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/HeadingComponent.cshtml)]
+
+For more information on HTML encoding with WebUtility, see [WebUtility.HtmlEncode](/dotnet/api/system.net.webutility.htmlencode?view=netframework-4.7.1#System_Net_WebUtility_HtmlEncode_System_String_).
```
SteveSandersonMS: Let's remove the HTML encoding stuff.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+Index page component (*Index.cshtml*):
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/Index.cshtml?start=1&end=11)]
+
+When the index page loads, the `HeadingComponent` loads. The component:
+
+* Displays a heading on the page.
+* Presents a form that allows the user to change the heading text and control its font style.
+
+Heading component (*HeadingComponent.cshtml*):
```
SteveSandersonMS: This is quite a big chunk of code to put in as the very first component that a reader sees. I wonder if it would be easier to grasp if it were simplified, e.g., just to a checkbox that toggles whether the text is italic. What do you think?

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+For more information on HTML encoding with WebUtility, see [WebUtility.HtmlEncode](/dotnet/api/system.net.webutility.htmlencode?view=netframework-4.7.1#System_Net_WebUtility_HtmlEncode_System_String_).
+
+## Razor directives
+
+A Razor file for Blazor specifies [Razor directives](https://docs.microsoft.com/aspnet/core/mvc/views/razor#directives) at the top of the file using the `@` symbol to distinguish Razor code from HTML markup. Razor directives supported by Blazor are shown in the table. `@model`, `@section`, and Tag Helper directives aren't supported at this time.
+
+| Directive | Description |
+| --------- | ----------- |
+| [@using](https://docs.microsoft.com/aspnet/core/mvc/views/razor#using) | Adds the C# `using` directive to the generated view. |
```
SteveSandersonMS: It's not a "view", it's a component class

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+| [@inherits](https://docs.microsoft.com/aspnet/core/mvc/views/razor#inherits) | Provides full control of the class that the component inherits. |
+| [@inject](https://docs.microsoft.com/aspnet/core/mvc/views/razor#inject) | Enables service injection from the [service container](https://docs.microsoft.com/aspnet/core/fundamentals/dependency-injection). For more information, see [Dependency injection into views](https://docs.microsoft.com/aspnet/core/mvc/views/dependency-injection). |
+| [@functions](https://docs.microsoft.com/aspnet/core/mvc/views/razor#functions) | Adds function-level content to a component. Functions include properties and methods accessible to the component. |
```
SteveSandersonMS: "function-level content" is a strange term. Replace with "a C# code block"?

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+*BlazorRocks.cshtml*:
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/BlazorRocks.cshtml?start=1&end=8)]
+
+*BlazorRocksBase.cs*:
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/BlazorRocksBase.cs)]
+
+The base class must derive from `BlazorComponent` (or implement `IComponent` directly) in the `Microsoft.AspNetCore.Blazor.Components` namespace for this approach to work.
```
SteveSandersonMS: Implementing `IComponent` isn't sufficient if you also want to derive a Razor file from it. It has to implement `BlazorComponent` for that.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+## Data binding
+
+### One-way databinding
```
SteveSandersonMS: As with earlier comments, I'd rather use the term "rendering" for this, because it avoids giving the wrong impression. Unless the code contains `bind`, there's no stateful link so I don't think we should use the term "databinding" for simple rendering.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+* The `HeadingFontStyle` property to the CSS property value for `font-style`.
+* `HeadingText` to the content of the `<h1>` element.
+
+```cshtml
+<h1 style="font-style:@HeadingFontStyle">@HeadingText</h1>
+```
+
+### Two-way databinding
+
+Two-way databinding to both components and DOM elements is accomplished with the `bind` attribute. The following example binds the `ItalicsCheck` property to the checkbox's `checked` attribute:
```
SteveSandersonMS: It's not bound to the checkbox's `checked` attribute; it's bound to the *checked state* of the checkbox. Toggling a checkbox doesn't change its HTML attributes.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+```cshtml
+<input type="checkbox" class="form-check-input" 
+    id="italicsCheck" bind="@ItalicsCheck">
+```
+
+Selecting the checkbox in the UI updates the property's value. Updating the property value in C# code updates the checkbox in the UI.
```
SteveSandersonMS: Selecting the checkbox -> Toggling the checkbox

> Updating the property value in C# code updates the checkbox in the UI.

No it doesn't. The checkbox is updated in the UI only when the component chooses to render itself. Since components render themselves after processing event handler code, this usually means that property updates are reflected in the UI immediately, but we should be clear that the UI is updated in response to *rendering*, not to changing .NET property values.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+**Format strings**
+
+Two-way databinding works with DateTime format stings (but not other format expressions at this time, such as currency or number formats):
```
SteveSandersonMS: Typo: stings -> strings

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+```cshtml
+<input type="date" bind="@StartDate" format="yyyy-MM-dd">
```
SteveSandersonMS: I *think* the syntax is `format-value` but please check it works!

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+```cshtml
+<div> ... </div>
+
+@functions {
+    public int Year { get; set; }
+    public Action<int> ValueChanged { get; set; }
+}
+```
+
+**User-defined binding**
```
SteveSandersonMS: Can we remove this section? We haven't decided whether user-defined binding defaults are going to be a thing in the long run. We need more examples of people having real use cases for it first, and I'd rather not give the impression that this is something typical devs should have reason to do. Thanks!

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+```cshtml
+<ul bind-foo="@SomeExpression" />
+```
+
+**Bind value and change handler attributes**
+
+`bind` can be specified with any value attribute name and change handler attribute name. The following example binds the current value of `SomeExpression` to `myvalue` and a change handler lambda to `myevent`:
+
+```cshtml
+<ul bind-myvalue-myevent="@SomeExpression">
```
SteveSandersonMS: That wouldn't do anything - could we give a realistic example instead?

I *think* that the following would work, but please verify if you are going to use it:

```
<input bind-oninput="@SomeExpression" /> <!-- Updates after each keystroke -->
```

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+    ...
+</ul>
+```
+
+## Event handling
+
+Blazor offers two event handling features, `@onclick` and `@onchange`. The following code calls the `UpdateHeading` method when the button is selected in the UI:
```
SteveSandersonMS: Hopefully we'll have @rynowak's real event handling syntax in before 0.2.0 so this section will change.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+The following code calls the `CheckboxChanged` method when the checkbox is changed in the UI:
+
+```cshtml
+<input type="checkbox" class="form-check-input" 
+    id="callsMethodCheck" @onchange(CheckboxChanged)>
+```
+
+`OnInit` and `OnInitAsync` execute code after the component has been initialized:
```
SteveSandersonMS: Are we missing a heading before this line? It seems odd to put information about lifecycle methods under the "binding" subheading.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+```csharp
+protected override void OnInit()
+{
+    ...
+}
+```
+
+```csharp
+protected override async Task OnInitAsync()
```
SteveSandersonMS: Could we specify why the two both exist? (i.e., if you want to perform an asynchronous operation, use `OnInitAsync` and use the `await` keyword in your code, otherwise simply use `OnInit`).

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+```csharp
+protected override async Task OnParametersSetAsync()
+{
+    ...
+}
+```
+
+`SetParameters` can be overridden to execute code before parameters are set:
```
SteveSandersonMS: Or, if you choose not to invoke `base.SetParameters`, your custom code can interpret the incoming `parameters` value in any way you need. For example, you don't have to assign the incoming `parameters` to properties on your class.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+{
+    var renderUI = true;
+
+    return renderUI;
+}
+```
+
+## Component disposal with IDisposable
+
+If a component implements [IDisposable](/dotnet/api/system.idisposable), the router disposes the component when the user navigates away from the component. A component can implement `IDisposable` in a C# base class that implements `BlazorComponent` (or `IComponent` directly). The following component uses `@implements IDisposable` and the `Dispose` method:
```
SteveSandersonMS: It's not the router that does this. It happens for *all* components when they get removed from the UI (e.g., if they were inside a `if (someBoolean) { ... }` block and `someBoolean` became `false` since the previous render), whether or not that removal was triggered by routing.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+@functions {
+    public void Dispose()
+    {
+        ...
+    }
+}
+```
+
+## Render child content
```
SteveSandersonMS: "child content" is a strange term, because regular HTML is also content.

Could this be changed to "child components"? Also in the sentence below.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+Blazor components can render child content with an element that matches the child component's name. The following Index page renders the content of the `HeadingComponent` (*HeadingComponent.cshtml*):
```
SteveSandersonMS: Phrasing is strange. Suggested:

Components can include child components simply by declaring them using HTML element syntax. The following Index page renders a `HeadingComponent` (*HeadingComponent.cshtml*) instance:

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+*ParentComponent.cshtml*:
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/ParentComponent.cshtml?start=1&end=7)]
+
+*ChildComponent.cshtml*:
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/ChildComponent.cshtml)]
+
+The body of the Bootstrap-styled panel is provided by `ChildContent`. `ChildContent` is bound to the `<ChildComponent>` element's content in the parent component by convention.
```
SteveSandersonMS: It's not bound; it just receives the content from inside the element in the parent.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+## Routing
+
+Routing in Blazor is achieved by providing a route template to each accessible component in the app. Route templates are converted into [RouteAttribute](/dotnet/api/microsoft.aspnetcore.mvc.routeattribute) implementations. Multiple route templates can be applied to a component. The following component responds to requests for `/BlazorRoute` and `/DifferentBlazorRoute`:
```
SteveSandersonMS: > Route templates are converted into [RouteAttribute](/dotnet/api/microsoft.aspnetcore.mvc.routeattribute) implementations.

That's a strange phrase. Suggestion:

When a `.cshtml` file with a `@page` directive is compiled, the resulting class is given a `[Route]` attribute specifying the route template. At runtime, the router looks for component classes with this attribute and renders whichever component has a route template matching the requested URL.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+Blazor components can receive and use route parameters from the route template provided in the `@page` directive:
+
+*RouteParameter.cshtml*:
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/RouteParameter.cshtml?start=1&end=8)]
+
+Not all route parameter constraints are supported at this time. Optional parameters aren't currently supported, so two `@page` directives are applied in the example above. The first permits navigation to the component without a parameter. The second `@page` directive takes the `{text}` route parameter and assigns the value to the `Text` property.
+
+For more information on attribute routing, see [Routing](https://docs.microsoft.com/aspnet/core/fundamentals/routing).
```
SteveSandersonMS: Let's not have this link, because it's specific to server-side routing.

docs/components/index.md
```diff
@@ -0,0 +1,315 @@
...
+
+## JavaScript/TypeScript interop
+
+To call JavaScript libraries or custom JavaScript/TypeScript code from .NET, the current approach is to register a named function with JavaScript/TypeScript. Place the `registerFunction` call in the *index.html* file or a JavaScript file (*\*.js*) loaded by the *index.html* file. Place the inline JavaScript or `<script>` tag below `<script type="blazor-boot"></script>`, and the JavaScript/TypeScript loads at the correct time and only executes once. 
+
+```javascript
+Blazor.registerFunction('doPrompt', message => {
```
SteveSandersonMS: For wider browser compatibility, can we avoid the lambda syntax? i.e., replace with:

```
Blazor.registerFunction('doPrompt', function(message) {
    return prompt(message);
});
```

guardrex:
@SteveSandersonMS Thanks ... I'll get right on this. Sorry you had to do so much work, but that's about what I expected given my two-days worth of experience with Blazor. I made an awful lot of *uneducated (incorrect) guesses*. :smile: lol

SteveSandersonMS:
@guardrex No problem at all, and thanks for writing all this! I appreciate it's hard to know what's going on in a totally new framework. Your time spent figuring this out is hopefully time that readers won't have to spend :)

docs/components/common/samples/2.x/ComponentsSample/Pages/HeadingComponent.cshtml
```diff
@@ -0,0 +1,63 @@
...
+    public bool ItalicsCheck { get; set; } = false;
+
+    // When UpdateHeading is executed:
+    //    HeadingText is assigned the HTML encoded value of 
+    //        NewHeadingText.
+    //    ItalicsCheck determines the value of HeadingFontStyle to 
+    //        set the font style of the heading.
+    public void UpdateHeading()
+    {
+        HeadingText = WebUtility.HtmlEncode(NewHeadingText);
```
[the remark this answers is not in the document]
guardrex: > Why call HtmlEncode on this? That shouldn't be necessary.

Bad example choice to provide context for directives (the example is providing a demo of `@using`).

I'll improve this by using `System.Globalization` and `ToTitleCase(string)`.

docs/components/index.md
guardrex: One thing that *LEAPS* right out is the link itself, which changes now that this is in the Blazor.Docs repo (that link was held over from when this PR sat in the aspnet/Docs repo). I'll fix it on this pass.

Anyway ... to your point ... This topic has a sample to accompany it. The dev can click this link (after I fix it) and get to the sample app. They take the ZIP of the docs repo to download the sample. Don't ask!! 🙈 lol ... We've had a lot of discussion about that, and we have a fix on the way to make sample downloading a snap. They run the sample the same way they would run any standalone Blazor app. [We probably need to address that further tho. Let's discuss this further on the second pass after the next commit.]

The sample matches most of what the topic covers. It also (the way I build them) is completely self-documenting. If a dev ONLY has the sample, they can still understand how the Blazor app works. When you run it, it shows the Razor code behind the component that makes the component work ... it shows that code in the sample's rendered output. This has been a very effective way to build samples when training devs on ASP.NET Core features (but I admit that my conclusion is based on my anecdotal experience talking to devs).

Now that we aren't going have the "Build your first" tutorial in the launch doc set, I think this sample app is even *MORE* important. There seems to be very few (or no) sample apps explained in detail (outside of your presentations/videos).

docs/components/index.md
guardrex: > I think the syntax is format-value but please check it works!

Yes, `format-value`. However, the `type="date"` must be removed.

docs/components/index.md
guardrex: > I think that the following would work

:cry: It didn't work. Tested with ...

```cshtml
<input bind-oninput="@InputProp" />

<h1>@InputProp</h1>
```

```csharp
public string InputProp { get; set; } = "xxx";
```
Changing the value of the input and clicking the **Update** button to force a component rendering.

I'll pull this for now ... let me know if I'm doing this wrong or if I can try something else.

guardrex:
@SteveSandersonMS That should get us closer.

Question: Use a *Pages* folder for components, or use a *Components* folder for components?

RE: `System.Globalization` and `ToTitleCase` ... I need something in the child component example (at the top) to demo a directive. We don't have to use an `@using` ... we could do an `@inject` (of something). I thought I would be able to use `CultureInfo` ... nope. Let's see what else we can come up with if you don't like that particular `@using`. It is a bit silly given that **it's title-casing a string**. 🤦‍♂️

RE: The BIG-ish example at the top of this topic ... Designed to provide context to most of the sections that follow. The example (and even the sample app) don't demo everything, but they do demo quite a bit. By providing code comments in the code examples, a dev should be able to get a feel for the basics. I did remove the heading text bits to cut it down tho as you recommended.

SteveSandersonMS:
> Question: Use a Pages folder for components, or use a Components folder for components?

The framework doesn't mind either way. I think the `Pages`/`Shared` split we have in the default templates makes some sense, but ultimately it's best if developers organize according to their own project shape. For example, grouping by user-visible feature or by area of business domain. One thing that makes no sense is grouping by file type, which lazy framework devs often do just because they have no other business domain.

> RE: System.Globalization and ToTitleCase ... I need something in the child component example (at the top) to demo a directive

ToTitleCase sounds good to me.

Thanks for the updates!

guardrex:
@danroth27 I made the updates to bind syntax, and I added a few bits from the 0.2.0 file.

danroth27:
@SteveSandersonMS Please take a look.

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+
+To download the sample:
+
+1. Download and unzip the [Blazor.Docs repository zip file](https://codeload.github.com/aspnet/Blazor.Docs/zip/master).
+1. Use the URL in the [sample link](https://github.com/aspnet/Docs/tree/master/aspnet/Blazor.Docs/docs/components/common/samples/) to navigate to the sample folder.
+1. See the [Get started](xref:client-side/blazor/get-started) topic for prerequisites.
+
+Blazor apps are built using *components*. A component is a self-contained chunk of user interface (UI), such as a page, dialog, or form. A component includes both the HTML markup to render along with the processing logic needed to inject data or respond to UI events. Typically, components are written using Razor files (*\*.cshtml*). Components are flexible and lightweight, and they can be nested, reused, and shared between projects.
+
+## Use of Razor
```
danroth27: Instead of introducing Razor as a separate concept that you need to first go off and familiarize yourself with, I think we should introduce Razor concepts as needed and then refer to the Razor docs for reference purposes. In fact, I'd prefer to not even mention the name "Razor" in this doc except as a single aside. In other words, I'd like to introduce Blazor component in way that makes sense to someone who has never used Razor before and doesn't require Razor knowledge as a prerequisite.

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+| [@inherits](https://docs.microsoft.com/aspnet/core/mvc/views/razor#inherits) | Provides full control of the class that the component inherits. |
+| [@inject](https://docs.microsoft.com/aspnet/core/mvc/views/razor#inject) | Enables service injection from the [service container](https://docs.microsoft.com/aspnet/core/fundamentals/dependency-injection). For more information, see [Dependency injection into views](https://docs.microsoft.com/aspnet/core/mvc/views/dependency-injection). |
+| `@layout` | Specifies a layout component. Layout components are used to avoid code duplication and inconsistency. |
+| [@page](https://docs.microsoft.com/aspnet/core/mvc/razor-pages#razor-pages) | Specifies that the component should handle requests directly. The `@page` directive can be specified with a route and optional parameters. Unlike Razor Pages, the `@page` directive doesn't need to be the first directive at the top of the file. |
+| [@using](https://docs.microsoft.com/aspnet/core/mvc/views/razor#using) | Adds the C# `using` directive to the generated component class. |
+
+## Component classes
```
danroth27: I like the heading for this section, but the first paragraph doesn't say anything about how Blazor components are implemented as .NET classes and it seems to go out of its way to immediately point out something that you can't do easily today.

I'd expect this section to cover:
- Components are implemented using classes
- Components are typically implemented in .cshtml files using a combination of HTML and C#
- Dynamic rendering logic (loops, conditionals, expressions, etc.) is added using an embedded C# syntax called Razor.
- The HTML markup and C# rendering logic are converted into a component class at build time. 
- The name of the generated .NET class matches the name of the file.
- Members of the component class are defined in a `@functions` block. In the `@functions` block you can specify component state (properties, fields) as well as methods for event handling or for defining other component logic.
- Component members can then be used as part of the component's rendering logic using C# expressions that start with `@` (partially covered by the Render C# expressions section already)

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+C# expressions are rendered by prefixing the `@` symbol. For example, a C# field is rendered by prefixing `@` to the field name. The following example evaluates and renders:
+
+* The `_HeadingFontStyle` field to the CSS property value for `font-style`.
+* `_HeadingText` to the content of the `<h1>` element.
+
+```cshtml
+<h1 style="font-style:@_HeadingFontStyle">@_HeadingText</h1>
+```
+
+## Two-way databinding
```
danroth27: I'd prefer to drop "Two-way" and just call this "Data binding". The problem is that two-way binding means different things in different UI frameworks. A bunch of folks expect "two-way binding" to mean that updating component state will trigger a render, which is not the case in Blazor. @SteveSandersonMS @rynowak

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+```csharp
+public static bool DoPrompt(string message)
+{
+    return RegisteredFunction.Invoke<bool>("doPrompt", message);
+}
+```
+
+This approach has the benefit of working with JavaScript build tools, such as [webpack](https://webpack.js.org/).
+
+The Mono team is working on a library that exposes standard browser APIs to .NET.
```
danroth27: I think we're missing sections on defining component parameters and also how to use components (i.e. `<Counter IncrementAmount="10" />`). Also, in the using components section I think we need to document that you have to use `@addTagHelper` to use a component that lives in a different assembly than the app assembly.

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+
+[!INCLUDE[](~/includes/blazor-preview-notice.md)]
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnet/Blazor.Docs/docs/components/common/samples/)
+
+To download the sample:
+
+1. Download and unzip the [Blazor.Docs repository zip file](https://codeload.github.com/aspnet/Blazor.Docs/zip/master).
+1. Use the URL in the [sample link](https://github.com/aspnet/Docs/tree/master/aspnet/Blazor.Docs/docs/components/common/samples/) to navigate to the sample folder.
+1. See the [Get started](xref:client-side/blazor/get-started) topic for prerequisites.
```
SteveSandersonMS: Could we put the "prerequisites" item first?

The phrasing of the second item is extremely strange. I don't know how to interpret what it's telling people to do. What sample link - the actual link whose text is "sample link", and if so, what is one meant to do with that URL? Could this be rephrased for clarity?

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+
+Razor is a syntax for embedding C# code into HTML markup files. Files containing Razor generally have a *\*.cshtml* file extension. We recommend that you become familiar with Razor while working with Blazor. For more information, see the [Razor syntax reference](https://docs.microsoft.com/aspnet/core/mvc/views/razor). Note that not all of the features of Razor are available in Blazor at this time.
+
+The following pair of Blazor components demonstrates how Blazor takes advantage of features in Razor, C#, and HTML. Many of these features are described in more detail later in this topic.
+
+Index page component (*Index.cshtml*):
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/Index.cshtml?start=1&end=11)]
+
+When the index page loads, the `HeadingComponent` loads. The component:
```
SteveSandersonMS: Probably fairer to say the `HeadingComponent` *executes*. Don't want to give the impression we're lazy-loading individual components.

docs/components/index.md
SteveSandersonMS: Agreed.

Can we remove the statements about what *isn't supported at this time*? It suggests we plan to add those things later, which is not the case for things we don't think fit into the programming model.

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+| `@layout` | Specifies a layout component. Layout components are used to avoid code duplication and inconsistency. |
+| [@page](https://docs.microsoft.com/aspnet/core/mvc/razor-pages#razor-pages) | Specifies that the component should handle requests directly. The `@page` directive can be specified with a route and optional parameters. Unlike Razor Pages, the `@page` directive doesn't need to be the first directive at the top of the file. |
+| [@using](https://docs.microsoft.com/aspnet/core/mvc/views/razor#using) | Adds the C# `using` directive to the generated component class. |
+
+## Component classes
+
+Blazor's Razor files (*\*.cshtml*) mix HTML markup and C# processing code in the same file, as shown in the sample `HeadingComponent` above. Blazor doesn't currently support `partial` component classes. However, the `@inherits` directive can be used to provide Blazor with a "code-behind" experience that separates view markup from processing code.
```
SteveSandersonMS: > Blazor doesn't currently support `partial` component classes. However, the `@inherits` directive can be used to provide Blazor with a "code-behind" experience that separates view markup from processing code.

I'd prefer to drop those sentences, because a newcomer would have no idea they even wanted to use `partial` classes anyway, so it's odd to be telling them they can't. We can still describe the "inherits" technique later as a more advanced topic.

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/BlazorRocksBase.cs)]
+
+The base class must derive from `BlazorComponent` in the `Microsoft.AspNetCore.Blazor.Components` namespace.
+
+### Render C# expressions
+
+C# expressions are rendered by prefixing the `@` symbol. For example, a C# field is rendered by prefixing `@` to the field name. The following example evaluates and renders:
+
+* The `_HeadingFontStyle` field to the CSS property value for `font-style`.
```
SteveSandersonMS: The "pascal case with leading underscore" variable name is a very strange naming convention. For normal C# code we go with `camelCase` for private fields (or `_camelCase` for private fields at a push), and `PascalCase` for public properties. No `_PascalCase`.

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+```cshtml
+<input type="checkbox" class="form-check-input" 
+    id="italicsCheck" bind="@ItalicsCheck">
+```
+
+When the check box is selected and cleared, the property's value is updated to `true` and `false`, respectively.
+
+The check box is updated in the UI only when the component is rendered, not in response to changing the property's value. Since components render themselves after event handler code executes, property updates are usually reflected in the UI immediately.
+
+Using `bind` with a `CurrentValue` (`<input bind="@CurrentValue" />`) is essentially equilivant to the following:
```
SteveSandersonMS: Typo on "equivalent".

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+
+**Format strings**
+
+Two-way databinding works with [DateTime](https://docs.microsoft.com/dotnet/api/system.datetime) format strings (but not other format expressions at this time, such as currency or number formats):
+
+```cshtml
+<input bind="@StartDate" format-value="yyyy-MM-dd" />
+
+@functions {
+    public DateTime StartDate { get; set; } =
```
SteveSandersonMS: Linebreak looks weird in the middle of this expression. Could just be one line.

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+
+@functions {
+    public DateTime StartDate { get; set; } = 
+        new DateTime(2020, 1, 1);
+}
+```
+
+**Conditional attributes**
+
+Blazor conditionally renders attributes based on the bound .NET value. If the value bound is `false` or `null`, Blazor won't render the attribute. If the value is `true`, the attribute is rendered minimized.
```
SteveSandersonMS: There's no binding here, so let's not use the term "bound". We just conditionally render attributes according to the specified value.

docs/components/index.md
guardrex: I think we should move this section to the top of the topic and let the much-shorter *BlazorRocks* example be the first example the readers see. The larger example of how to change the heading style is still available in the sample and still used in the code snippets below. On the next commit, let's take a look and see if it works better.

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+
+@functions {
+    public int Year { get; set; }
+    public Action<int> ValueChanged { get; set; }
+}
+```
+
+## Event handling
+
+Blazor offers two event handling features, `onclick` and `onchange`. The following code calls the `UpdateHeading` method when the button is selected in the UI:
```
SteveSandersonMS: We now handle arbitrary events, not just `onclick` and `onchange`. In general the rule is: if you have an attribute named `onX`, and supply a delegate-typed value, then we treat it as an event handler for the event named `X`. For example, `onclick`, `onsubmit`, etc.

An easier way to explain this is just to say that to handle an event, developers assign a value to an attribute named after the event. The attribute's name must always start with `on`.

SteveSandersonMS: Also the word "selected" should be "clicked".

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+{
+    var renderUI = true;
+
+    return renderUI;
+}
+```
+
+## Component disposal with IDisposable
+
+If a component implements [IDisposable](/dotnet/api/system.idisposable), the [Dispose method](https://docs.microsoft.com/dotnet/standard/garbage-collection/implementing-dispose) is called to release unmanaged resources when the component is removed from the UI. A component can implement `IDisposable` in a C# base class that implements `BlazorComponent`. The following component uses `@implements IDisposable` and the `Dispose` method:
```
SteveSandersonMS: Not a fan of the phrase "release unmanaged resources". I know that's used elsewhere in .NET docs but it's wrong. Disposal is a more general programming pattern than that; you can release arbitrary things or perform other teardown logic inside the dispose call. Can we just say "dispose is called when the component is removed from the UI"?

SteveSandersonMS: > A component can implement `IDisposable` in a C# base class that implements `BlazorComponent`.

That's a bit of a non-sequitor, as it doesn't relate to the subsequent code sample. Can we remove that sentence?

docs/components/index.md
```diff
@@ -0,0 +1,360 @@
...
+
+*ParentComponent.cshtml*:
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/ParentComponent.cshtml?start=1&end=7)]
+
+*ChildComponent.cshtml*:
+
+[!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/ChildComponent.cshtml)]
+
+The body of the Bootstrap-styled panel is provided by `ChildContent`. `ChildContent` is loaded where the `<ChildComponent>` element appears in the parent component.
```
SteveSandersonMS: Can we replace "loaded" with "rendered" for better precision?

docs/components/index.md
guardrex: Perhaps so ... but there are larger forces at work here. Later, we must use this language on pain-of-death ☠️ :smile:. This is the text we use on the aspnet/Docs side. https://github.com/aspnet/Docs/blame/master/aspnetcore/tutorials/index.md#L84

I wrote detailed instructions on the .NET Core docs side on https://github.com/dotnet/docs/pull/1832 ... that are live at https://docs.microsoft.com/dotnet/samples-and-tutorials/#viewing-and-downloading-samples.

Because we're in a separate repo right now, yes, I'll institute the detailed instructions elsewhere (docs index page or tutorials index page) along the lines of how the .NET Core docs handle it and link it via a "how to download" link here. When this content moves into the aspnet/Docs, we'll have to drop it tho and go with the direction that aspnet/Docs takes.

docs/components/index.md
guardrex: The first sentence explains it. However, the second sentence could be dropped. I disagree on removing the entire paragraph because this topic isn't only for newcomers. An obvious question is going to be how to get the component logic out of the component markup. It was one of the first questions I had when looking at Blazor component examples, and I learned these details in @rstropek's content. @danroth27 Thoughts on this?

docs/components/index.md
guardrex: Yes, of course. It was a property that we changed to a field on an earlier commit. I shot the underscore without lower-casing.

docs/components/index.md
danroth27: I would prefer we predominantly show components implemented inline and not using the separate base class approach. We may eventually decide to support a partial class model, but for now the recommended practice is to implement components inline in a single .cshtml file.

docs/components/index.md
guardrex: @danroth27 

> I would prefer we predominantly show components implemented inline and not using the separate base class approach. We may eventually decide to support a partial class model, but for now the recommended practice is to implement components inline in a single .cshtml file.

This is the only place it's shown, and it isn't shown as part of any other example. I suggest that with only one example (btw- I moved this section to the bottom of the topic) that the topic (and the sample) predominantly shows inline. Do you want the section completely removed tho? I will drop the 2nd sentence for sure.

danroth27: > An obvious question is going to be how to get the component logic out of the component markup. It was one of the first questions I had when looking at Blazor component examples, and I learned these details in @rstropek's content. @danroth27 Thoughts on this?

For now we want the focus to be on implementing components inline. We can have a separate section later in the doc that shows how to separate out the logic using `@inherits`, but it should be the exception, not the rule in this doc. That said, I'm hearing a steady stream of feedback that folks prefer the C# logic to be in a separate file, so eventually I believe we will need to support partial classes.

guardrex: Ok ... pulling this section for now (and pulling from the sample app, too).

docs/components/index.md
guardrex: We must say "select"/"selected" given that Blazor apps run on mobile devices and tablets (no mouse) ... it's MS style manual guidance.

docs/components/index.md
guardrex: The section is there, but I called them "route parameters" (`## Route parameters`) ... shall I change the name to "component parameters"?

SteveSandersonMS: There's a difference between the two:

* Component parameters are what you can pass in using HTML attribute syntax, e.g., `<MyComponent SomeValue=@Something />`
* Route parameters are the tokens in a route pattern, e.g., `@page "/products/{productId}"`.

The router uses route parameters to populate the corresponding component parameters. So they are related, but different concepts.

guardrex:
@danroth27 @SteveSandersonMS Ok ... see where the last commit gets us.

docs/components/index.md
```diff
@@ -0,0 +1,337 @@
...
+---
+# Blazor components
+
+By [Luke Latham](https://github.com/guardrex)
+
+[!INCLUDE[](~/includes/blazor-preview-notice.md)]
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnet/Blazor.Docs/docs/components/common/samples/) ([how to download](xref:client-side/blazor/index#view-and-download-samples)). See the [Get started](xref:client-side/blazor/get-started) topic for prerequisites.
+
+Blazor apps are built using *components*. A component is a self-contained chunk of user interface (UI), such as a page, dialog, or form. A component includes both the HTML markup to render along with the processing logic needed to inject data or respond to UI events. Typically, components are written using Razor files (*\*.cshtml*). Components are flexible and lightweight, and they can be nested, reused, and shared between projects.
```
danroth27: ~~Typically, components are written using Razor files (*\*.cshtml*)~~

docs/components/index.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Component classes
+
+Components are implemented using C# classes. Classes are usually saved in *\*.cshtml* files using a combination of HTML and C#.
```
danroth27: How about we replace the first three paragraphs in this section with:

Blazor components are typically implemented in *\*.cshtml* files using a combination of C# and HTML markup. The UI for the component is defined using HTML. Dynamic rendering logic (for example, loops, conditionals, expressions) is added using an embedded C# syntax called [Razor](https://docs.microsoft.com/aspnet/core/mvc/views/razor). When a Blazor app is compiled, the HTML markup and C# rendering logic are converted into a component class. The name of the generated class matches the name of the file.

Members of the component class . . .

docs/components/index.md
```diff
@@ -0,0 +1,337 @@
...
+
+@functions {
+    private string _headingFontStyle = "italic";
+    private string _headingText = "Put on your new Blazor!";
+}
+```
+
+## Directives
+
+A component may specify one or more [directives](https://docs.microsoft.com/aspnet/core/mvc/views/razor#directives) at the top of the file.
```
danroth27: Razor directives don't actually have to be at the top of the file.

Replace with: Blazor supports the following Razor directives when authoring components:

docs/components/index.md
```diff
@@ -0,0 +1,337 @@
...
+```cshtml
+<h1 style="font-style:@_headingFontStyle">@_headingText</h1>
+
+@functions {
+    private string _headingFontStyle = "italic";
+    private string _headingText = "Put on your new Blazor!";
+}
+```
+
+## Directives
```
danroth27: How about we rename the "Additional resources" section at the end of this doc to "Razor support" and move this table into that section along with the link to the Razor syntax reference doc.

docs/components/index.md
```diff
@@ -0,0 +1,337 @@
...
+| [@inject](https://docs.microsoft.com/aspnet/core/mvc/views/razor#inject) | Enables service injection from the [service container](https://docs.microsoft.com/aspnet/core/fundamentals/dependency-injection). For more information, see [Dependency injection into views](https://docs.microsoft.com/aspnet/core/mvc/views/dependency-injection). |
+| `@layout` | Specifies a layout component. Layout components are used to avoid code duplication and inconsistency. |
+| [@page](https://docs.microsoft.com/aspnet/core/mvc/razor-pages#razor-pages) | Specifies that the component should handle requests directly. The `@page` directive can be specified with a route and optional parameters. Unlike Razor Pages, the `@page` directive doesn't need to be the first directive at the top of the file. |
+| [@using](https://docs.microsoft.com/aspnet/core/mvc/views/razor#using) | Adds the C# `using` directive to the generated component class. |
+| [@addTagHelper](xref:mvc/views/razor#tag-helpers) | Use `@addTagHelper` to use a component in a different assembly than the app's assembly. |
+
+## Databinding
```
danroth27: I think it's two words: [Data binding](https://en.wikipedia.org/wiki/Data_binding)

docs/components/index.md
```diff
@@ -0,0 +1,337 @@
...
+
+```cshtml
+<h1 style="font-style:@_headingFontStyle">@_headingText</h1>
+
+@functions {
+    private string _headingFontStyle = "italic";
+    private string _headingText = "Put on your new Blazor!";
+}
+```
+
```
danroth27: Add the following text at the end of this section:

After the component is initially rendered the component regenerates its render tree in response to events. Blazor then compares the new render tree against the previous one and applies any modifications to the browser DOM.

docs/components/index.md
danroth27: Before we get into data binding, I think we should have a section here called "Using components".

## Using components

Once you define a component it can be used to implement other components. The markup for using a component looks like an HTML tag where the name of the tag is the component type.

```html
<Counter />
```

Components can also have parameters, which you define using public properties on the component class. Use attributes to specify arguments for a component in markup.

```html
<Counter IncrementAmount="10" />
```

docs/components/index.md
guardrex: I've had trouble a couple of times in testing when, for example, a directive is preceded by a Razor comment (or HTML comment). It might have been a one-off thing tho. It may have also just been a RP scenario, where RP might be a bit fussier about it.

docs/components/index.md
guardrex: This is here to explain Blazor top-to-bottom matching the layout of a Razor file. The first thing the dev sees looking that these files are the directives. Therefore, I felt that explaining them first (briefly) matched the flow of what a dev sees.

We could have a short directives paragraph here (perhaps just two sentences + a bookmark link) and move the table to the bottom. Do you want to try that approach?

docs/components/index.md
guardrex: We have the *Child components and component parameters* section. On the next commit, I'll take your language :point_up:, inject/revise that section with it, and move it above the *Data binding* section.

guardrex:
@danroth27 Updates made. I left a short blurb and bookmark link to the *Razor support* section so that we maintain some coverage for directives early. See if that works.

docs/components/index.md
danroth27: The thing is, you don't need any directives initial besides `@functions` to start writing a component. The directives come later for specific scenarios that are covered in their respective sections (DI, usings, routing, etc.)

docs/components/index.md
```diff
@@ -0,0 +1,337 @@
...
+After the component is initially rendered, the component regenerates its render tree in response to events. Blazor then compares the new render tree against the previous one and applies any modifications to the browser's Document Object Model (DOM).
+
+## Child components and component parameters
```
danroth27: I think this section should be broken into three: Using components, Component parameters, Child content.

docs/components/index.md
guardrex: > you don't need any directives ... to start writing a component

... but this is a reference topic . What we usually try to do 😅 in these types of topics is write from general-to-specific and top-to-bottom. This arrangement flips that. We now have general coverage (directives are the most general Razor configuration) at the bottom and top-of-the-file coverage (the first thing a dev sees) at the end of the topic.

I pulled it at the top. I'm happy we're not cutting it entirely. I'm sure devs will see it when they get to the bottom.

guardrex:
Must be my local .gitignore is lacking it and it snuck into one of my commits. I'll clear it out on the next one.

guardrex:
@danroth27 I broke that section and tried to stay true to the sample app. We can mod the sample and split it there too, if needed. I use line highlighting to help the reader follow along in the code snippets.

docs/components/index.md
```diff
@@ -0,0 +1,351 @@
...
+
+```cshtml
+<input bind="@StartDate" format-value="yyyy-MM-dd" />
+
+@functions {
+    public DateTime StartDate { get; set; } = new DateTime(2020, 1, 1);
+}
+```
+
+**Conditional attributes**
```
danroth27: This should really go in the "Razor support" section as it doesn't really have anything to do with data binding. It's just a Razor feature that we support.

guardrex:
@SteveSandersonMS Do you have additional changes?

docs/components/index.md
```diff
@@ -0,0 +1,355 @@
...
+
+## Razor support
+
+**Razor directives**
+
+Razor directives active with Blazor apps are shown in the following table.
+
+| Directive | Description |
+| --------- | ----------- |
+| [@functions](https://docs.microsoft.com/aspnet/core/mvc/views/razor#functions) | Adds a C# code block to a component. |
```
SteveSandersonMS: Inconsistent formatting in the "Directive" column. Some of the items are surrounded by backticks; some aren't. I think all of them should be.

PR #14, opened 2018-04-03
Add chapter about dependency injection in Blazor

### LINKED ISSUE #6: Dependency injection in Blazor topic
Title: Dependency injection in Blazor
UID: client-side/blazor/dependency-injection
URL: https://docs.microsoft.com/aspnet/core/client-side/blazor/dependency-injection

* Configuring services
* Resolving services
* Built-in services

@rstropek will take this one. Thank you @rstropek! :rocket:

rstropek:
Fixes #6

This pull request adds a chapter about DI in Blazor as discussed with @guardrex. Compared to the version he reviewed, I added a code sample at the end of the document showing how to implement `IDisposable` in a Blazor component.

### DIFF
--- a/docs/dependency-injection.md
+++ b/docs/dependency-injection.md
@@ -44,7 +44,7 @@ static void Main(string[] args)
 }
 ```
 
-`BrowserServiceProvider` receives an action with which you can add your app services to DI. `services` references the underlying `IServiceCollection`, which is a list of service descriptor objects ([Microsoft.Extensions.DependencyInjection.ServiceDescriptor](https://docs.microsoft.com/dotnet/api/microsoft.extensions.dependencyinjection.servicedescriptor)). Services are added by providing service descriptors to the service collection. The following is a code sample demonstrating the concept:
+`BrowserServiceProvider` receives an action where app services are added to DI. `services` references the underlying `IServiceCollection`, which is a list of service descriptor objects ([Microsoft.Extensions.DependencyInjection.ServiceDescriptor](https://docs.microsoft.com/dotnet/api/microsoft.extensions.dependencyinjection.servicedescriptor)). Services are added by providing service descriptors to the service collection. The following is a code sample demonstrating the concept:
 
 ```csharp
 @using Microsoft.Extensions.DependencyInjection
--- a/docs/dependency-injection.md
+++ b/docs/dependency-injection.md
@@ -35,9 +35,9 @@ After creating a new app, examine the `Main` method in *Program.cs*:
 ```csharp
 static void Main(string[] args)
 {
-    var serviceProvider = new BrowserServiceProvider(configure =>
+    var serviceProvider = new BrowserServiceProvider(services =>
     {
-        // Add any custom services here
+        // Add custom services here
     });
 
     new BrowserRenderer(serviceProvider).AddComponent<App>("app");
@@ -51,9 +51,9 @@ static void Main(string[] args)
 
 static void Main(string[] args)
 {
-    var serviceProvider = new BrowserServiceProvider(configure =>
+    var serviceProvider = new BrowserServiceProvider(services =>
     {
-        configure.AddSingleton<IDataAccess, DataAccess>();
+        services.AddSingleton<IDataAccess, DataAccess>();
     });
 
     new BrowserRenderer(serviceProvider).AddComponent<App>("app");
@@ -168,7 +168,7 @@ Note the following prerequisites for constructor injection:
 
 ## Service lifetime
 
-Note that Blazor doesn't automatically dispose injected services that implement `IDisposable`. Components can implement `IDisposable`. Components are disposed when the user navigates away from the component, then the component can dispose any transient services.
+Note that Blazor doesn't automatically dispose injected services that implement `IDisposable`. Components can implement `IDisposable`. Components and their transient services are disposed when the user navigates away from the component.
 
 The following code sample demonstrates how to implement `IDisposable` in a component:
 
--- a/docs/dependency-injection.md
+++ b/docs/dependency-injection.md
@@ -166,28 +166,6 @@ Note the following prerequisites for constructor injection:
 * The applicable constructor must be *public*.
 * There must only be one applicable constructor. In case of an ambiguity, DI throws an exception.
 
-## Service lifetime
-
-Note that Blazor doesn't automatically dispose injected services that implement `IDisposable`. Components can implement `IDisposable`. Components and their transient services are disposed when the user navigates away from the component.
-
-The following code sample demonstrates how to implement `IDisposable` in a component:
-
-```csharp
-...
-@using System;
-@implements IDisposable
-...
-
-@functions {
-    ...
-    public void Dispose()
-    {
-        // Add code for disposing here
-        ...
-    }
-}
-```
-
 ## Additional resources
 
 * [Dependency injection in ASP.NET Core](https://docs.microsoft.com/aspnet/core/fundamentals/dependency-injection)
--- a/docs/dependency-injection.md
+++ b/docs/dependency-injection.md
@@ -44,7 +44,7 @@ static void Main(string[] args)
 }
 ```
 
-`BrowserServiceProvider` receives an action where app services are added to DI. `services` references the underlying `IServiceCollection`, which is a list of service descriptor objects ([Microsoft.Extensions.DependencyInjection.ServiceDescriptor](https://docs.microsoft.com/dotnet/api/microsoft.extensions.dependencyinjection.servicedescriptor)). Services are added by providing service descriptors to the service collection. The following is a code sample demonstrating the concept:
+`BrowserServiceProvider` receives an action where app services are added to DI. `services` references the underlying `IServiceCollection`, which is a list of service descriptor objects ([Microsoft.Extensions.DependencyInjection.ServiceDescriptor](https://docs.microsoft.com/dotnet/api/microsoft.extensions.dependencyinjection.servicedescriptor)). Services are added by providing service descriptors to the service collection. The following code sample demonstrates the concept:
 
 ```csharp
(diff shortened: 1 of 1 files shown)


docs/dependency-injection.md
```diff
@@ -0,0 +1,190 @@
...
+---
+# Dependency injection in Blazor
+
+By [Rainer Stropek](https://www.timecockpit.com)
+
+[!INCLUDE[](~/includes/blazor-preview-notice.md)]
+
+Blazor has [dependency injection (DI)](https://docs.microsoft.com/aspnet/core/fundamentals/dependency-injection) built-in. Blazor apps can use built-in services by having them injected into components. Blazor apps can also define custom services and make them available via DI.
+
+## What is dependency injection?
```
SteveSandersonMS: Firstly, I really like the overall tone of voice here and think this is just the right amount of background detail. Great job!

Secondly, I *slightly* object to the idea that "component should *not* create an instance of `DataAccess`" as if that were always true. That's entirely the choice of the application developer. Sometimes there's value in working with an abstraction; sometimes it would just be unnecessary friction. It depends on the scenario. The DI system in Blazor is equally happy to supply instances of concrete service classes as it is to deal with interfaces.

I'd say something like:

DI is a technique for accessing services configured in a central location. This can be useful if:

* You want to share a single instance of a service class across many components (known as a "singleton" service)
* Or, if you want to decouple components from particular concrete service classes and only reference abstract interfaces. For example, you might have an interface `IDataAccess` implemented by a concrete class `DataAccess`. If your components use DI to receive an `IDataAccess` implementation, they aren't coupled to any concrete type. That means the implementation could easily be swapped, perhaps to a mock implementation in unit tests.

Blazor's DI system is responsible for supplying instances of services to components. It also resolves dependencies recursively, so that services themselves can depend on further services, and so on. DI is configured during startup of the app. An example is shown later in this topic.

docs/dependency-injection.md
```diff
@@ -0,0 +1,190 @@
...
+
+Blazor's DI system is responsible for connecting component abstractions with their concrete implementations. To connect the abstractions with the implementations, DI is configured during startup of the app. An example is shown later in this topic.
+
+## Use of existing .NET mechanisms
+
+DI in Blazor is based on .NET's [System.IServiceProvider](https://docs.microsoft.com/dotnet/api/system.iserviceprovider) interface. It defines a generic mechanism for retrieving a service object in .NET apps.
+
+Blazor's implementation of `System.IServiceProvider` obtains its services from an underlying [Microsoft.Extensions.DependencyInjection.IServiceCollection](https://docs.microsoft.com/dotnet/api/microsoft.extensions.dependencyinjection.iservicecollection).
+
+## Add services to DI
```
SteveSandersonMS: Cool - good explanation about configuration!

docs/dependency-injection.md
```diff
@@ -0,0 +1,190 @@
...
+| `Scoped`    | Blazor doesn't currently have the concept of DI scopes. `Scoped` behaves like `Singleton`. Therefore, prefer `Singleton` and avoid `Scoped`. |
+
+## System services
+
+Blazor provides default services that are automatically added to the service collection of a Blazor app. The following table shows a list of the default services currently provided by Blazor.
+
+| Method       | Description |
+| ------------ | ----------- |
+| `IUriHelper` | Helpers for working with URIs and navigation state (singleton). |
+| `HttpClient` | Provides methods for sending HTTP requests and receiving HTTP responses from a resource identified by a URI (singleton). Note that this instance of [System.Net.Http.HttpClient](https://docs.microsoft.com/dotnet/api/system.net.http.httpclient) is using the browser for handling the HTTP traffic in the background. Its [BaseAddress](https://docs.microsoft.com/dotnet/api/system.net.http.httpclient.baseaddress) is automatically set to the base URI prefix of the Blazor app. |
```
SteveSandersonMS: In a couple of places it says "is using" or "is referencing". Maybe it's a language difference, but a more common phrasing would be simply "uses" or "references". Not a big deal though!

docs/dependency-injection.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Request a service in a component
```
SteveSandersonMS: I like this section - well explained!

docs/dependency-injection.md
```diff
@@ -0,0 +1,190 @@
...
+@page "/demo"
+@inherits ComponentBases
+
+<h1>...</h1>
+...
+```
+
+## Dependency injection in services
+
+Complex services might require additional services. In the prior example, `DataAccess` might require Blazor's default service `HttpClient`. `@inject` or the `InjectAttribute` can't be used in services. *Constructor injection* must be used instead. Require services are added by adding parameters to the service's constructor. When dependency injection creates the service, it recognizes the services it requires in the constructor and provides them accordingly.
```
SteveSandersonMS: Typo: "Require services" should be "Required services"

docs/dependency-injection.md
```diff
@@ -0,0 +1,190 @@
...
+
+| Method      | Description |
+| ----------- | ----------- |
+| `Singleton` | Blazor creates a *single instance* of your service. All components requiring this service receive a reference to this instance. |
+| `Transient` | Whenever a component requires this service, it receives a *new instance* of the service. |
+| `Scoped`    | Blazor doesn't currently have the concept of DI scopes. `Scoped` behaves like `Singleton`. Therefore, prefer `Singleton` and avoid `Scoped`. |
+
+## System services
+
+Blazor provides default services that are automatically added to the service collection of a Blazor app. The following table shows a list of the default services currently provided by Blazor.
```
SteveSandersonMS: I wonder if it's worth clarifying that these are the default services in a `BrowserServiceProvider` (which is what the project template uses by default in `Program.cs`). If someone creates their own different service provider, it wouldn't contain these services.

guardrex:
I only see a handful of very minor grammatical changes to make. To save time, I'm just going to shoot them in there and then sign off. Stand by ....

guardrex:
Fixed everything outside of the *custom service provider with the default services* question.

docs/dependency-injection.md
```diff
@@ -0,0 +1,193 @@
...
+Blazor's DI system is responsible for supplying instances of services to components. DI also resolves dependencies recursively so that services themselves can depend on further services. DI is configured during startup of the app. An example is shown later in this topic.
+
+## Add services to DI
+
+After creating a new app, examine the `Main` method in *Program.cs*:
+
+```csharp
+static void Main(string[] args)
+{
+    var serviceProvider = new BrowserServiceProvider(configure =>
```
danroth27: One more thing: I'm proposing that rename `configure` to `services` so that it's consistent with ASP.NET Core.

danroth27: https://github.com/aspnet/Blazor/pull/494

guardrex:
@danroth27 One little thing tho that I find strange about this one ...

> Components are disposed when the user navigates away from the component, then the component can dispose any transient services.

If the component is disposed when the user navigates away ... then how can it dispose transient services? ... It reads like it's doing something *after* it has been destroyed.

Can we restructure it? ... e.g. ...

> Components and their transient services are disposed when the user navigates away from the component.

danroth27:
The more I think about this less sure I am that having disposable transient services is a thing. You didn't create the service, so it's weird that you would be responsible for disposing it. @pakrym What's your thoughts on this?

docs/dependency-injection.md
```diff
@@ -0,0 +1,193 @@
...
+## Default services
+
+Blazor provides default services that are automatically added to the service collection of an app. The following table shows a list of the default services currently provided by Blazor's `BrowserServiceProvider`.
+
+| Method       | Description |
+| ------------ | ----------- |
+| `IUriHelper` | Helpers for working with URIs and navigation state (singleton). |
+| `HttpClient` | Provides methods for sending HTTP requests and receiving HTTP responses from a resource identified by a URI (singleton). Note that this instance of [System.Net.Http.HttpClient](https://docs.microsoft.com/dotnet/api/system.net.http.httpclient) uses the browser for handling the HTTP traffic in the background. Its [BaseAddress](https://docs.microsoft.com/dotnet/api/system.net.http.httpclient.baseaddress) is automatically set to the base URI prefix of the app. |
+
+Note that it is possible to use a custom services provider instead of the default `BrowserServiceProvider` that's added by the default template. A custom service provider doesn't automatically provide the default services listed in the table. Those services must be added to the new service provider explicitly.
```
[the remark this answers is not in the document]
SteveSandersonMS: That’s not something we have the option to do. If the developer provides their own IServiceProvider we can’t control what it returns; it’s up to them to delegate calls to a BrowserServiceProvider if they want to.

SteveSandersonMS:
@danroth27 That requirement is described in https://github.com/aspnet/Blazor/issues/464

pakrym:
> The more I think about this less sure I am that having disposable transient services is a thing. You didn't create the service, so it's weird that you would be responsible for disposing it.

You don't usually create singleton services, transients are created and should be disposed by DI container.

docs/dependency-injection.md
```diff
@@ -0,0 +1,193 @@
...
+}
+```
+
+Note the following prerequisites for constructor injection:
+
+* There must be one constructor whose arguments can all be fulfilled by dependency injection. Note that additional parameters not covered by DI are allowed if default values are specified for them.
+* The applicable constructor must be *public*.
+* There must only be one applicable constructor. In case of an ambiguity, DI throws an exception.
+
+## Service lifetime
```
danroth27: Services are created by the service provider, so they need to be disposed by the service provider. We need to address this as part of https://github.com/aspnet/Blazor/issues/464, but regardless I think this section should be removed as it contradicts that guidance. The content about implementing `IDisposable` on components should be covered in the components doc, not here.

PR #17, opened 2018-04-04
Add layout chapter for Blazor docs

### LINKED ISSUE #7: Blazor layouts topic
Title: Blazor layouts
UID: client-side/blazor/layouts
URL: https://docs.microsoft.com/aspnet/core/client-side/blazor/layouts

* Defining layouts
* Using layouts

@rstropek will take this one. Thank you @rstropek! :rocket:

rstropek:
Fixes #7 

Adds a docs chapter about Blazor layouts as discussed with @guardrex

### DIFF
--- a/docs/layouts.md
+++ b/docs/layouts.md
@@ -17,14 +17,14 @@ By [Rainer Stropek](https://www.timecockpit.com)
 
 [!INCLUDE[](~/includes/blazor-preview-notice.md)]
 
-Blazor apps typically contain more than one page. Layout elements, such as menus, copyright messages, and logos, must be present on all pages. Copying the code of these layout elements onto all of the pages of an app isn't an efficient solution. Such duplication is hard to maintain and probably leads to inconsistent content over time. *Blazor layouts* solve this problem.
+Blazor apps typically contain more than one page. Layout elements, such as menus, copyright messages, and logos, must be present on all pages. Copying the code of these layout elements onto all of the pages of an app isn't an efficient solution. Such duplication is hard to maintain and probably leads to inconsistent content over time. *Layouts* solve this problem.
 
-## What are Blazor layouts?
+## What are layouts?
 
-Technically, a layout is just another Blazor component. A layout is defined in a Razor template or in C# code and can contain data binding, dependency injection, and other ordinary features of components. Two additional aspects are special and turn a Blazor component into a Blazor layout:
+Technically, a layout is just another Blazor component. A layout is defined in a Razor template or in C# code and can contain data binding, dependency injection, and other ordinary features of components. Two additional aspects are special and turn a *component* into a *layout*:
 
 * The layout component must implement `Microsoft.AspNetCore.Blazor.Layouts.ILayoutComponent`. This interface adds a property `Body` to the component, which contains the content to be rendered inside the layout.
-* The layout component contains the Razor keyword `@Body`. During rendering, it's replaced by the content of the layout.
+* The layout component uses the `Body` property to specify where the body content should be rendered. This is done using the Razor syntax `@Body`. During rendering, it's replaced by the content of the layout.
 
 The following code sample shows the Razor template of a layout component. Note the use of `ILayoutComponent` and `@Body`:
 
@@ -33,7 +33,6 @@ The following code sample shows the Razor template of a layout component. Note t
 
 <header>
     <h1>ERP Master 3000</h1>
-    <p>Current user: @UserName</p>
 </header>
 
 <nav>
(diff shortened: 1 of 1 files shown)


docs/layouts.md
```diff
@@ -0,0 +1,130 @@
+---
+title: Blazor layouts
```
SteveSandersonMS: As a general point, I think it would be good if we didn't repeat the word "Blazor" so much. Since this is the Blazor documentation, that context is already established. Could we just say "Layouts" everywhere in this doc rather than branding them as "Blazor layouts"? Thanks 😄

docs/layouts.md
```diff
@@ -0,0 +1,130 @@
...
+---
+# Blazor layouts
+
+By [Rainer Stropek](https://www.timecockpit.com)
+
+[!INCLUDE[](~/includes/blazor-preview-notice.md)]
+
+Blazor apps typically contain more than one page. Layout elements, such as menus, copyright messages, and logos, must be present on all pages. Copying the code of these layout elements onto all of the pages of an app isn't an efficient solution. Such duplication is hard to maintain and probably leads to inconsistent content over time. *Blazor layouts* solve this problem.
+
+## What are Blazor layouts?
```
SteveSandersonMS: This is a superb explanation - great job!

docs/layouts.md
```diff
@@ -0,0 +1,130 @@
...
+* The layout component contains the Razor keyword `@Body`. During rendering, it's replaced by the content of the layout.
+
+The following code sample shows the Razor template of a layout component. Note the use of `ILayoutComponent` and `@Body`:
+
+```csharp
+@implements ILayoutComponent
+
+<header>
+    <h1>ERP Master 3000</h1>
+    <p>Current user: @UserName</p>
```
SteveSandersonMS: It might be better not to have the `@UserName` here (or the property below) because it raises the question of "How do I populate the `UserName` value?" and this document doesn't answer that. Nor do I think there is a good answer right now.

docs/layouts.md
guardrex: @rstropek ... "Blazor" should stay in the title and the H1:

* All of the doc titles must say "Blazor" somewhere for SEO.
* The H1 should mirror the title. It's not a MS style manual requirement, but that's what we do in the aspnet/Docs repo. We're carrying that over in this doc set.
* Having said that tho, the title doesn't need to read *Blazor layouts*. Other options include: *Layouts in Blazor*, *Use layouts to avoid markup and code duplication in Blazor*, etc. ... so long as "Blazor" is in there somewhere. `@danroth27` may have a preference tho. This one, *Blazor layouts*, was approved in our high-level planning.

Regarding the rest of the references, yes, sounds good ... "layouts" unless you need to distinguish from other non-Blazor layouts.

docs/layouts.md
rstropek: Great feedback, thank you. I will change that in the course of the day.

docs/layouts.md
rstropek: Ok. Will leave Blazor in the title and remove it in the rest of the text where appropriately.

docs/layouts.md
```diff
@@ -0,0 +1,130 @@
...
+
+[!INCLUDE[](~/includes/blazor-preview-notice.md)]
+
+Blazor apps typically contain more than one page. Layout elements, such as menus, copyright messages, and logos, must be present on all pages. Copying the code of these layout elements onto all of the pages of an app isn't an efficient solution. Such duplication is hard to maintain and probably leads to inconsistent content over time. *Blazor layouts* solve this problem.
+
+## What are Blazor layouts?
+
+Technically, a layout is just another Blazor component. A layout is defined in a Razor template or in C# code and can contain data binding, dependency injection, and other ordinary features of components. Two additional aspects are special and turn a Blazor component into a Blazor layout:
+
+* The layout component must implement `Microsoft.AspNetCore.Blazor.Layouts.ILayoutComponent`. This interface adds a property `Body` to the component, which contains the content to be rendered inside the layout.
```
danroth27: @guardrex can we make this an API ref link?

docs/layouts.md
```diff
@@ -0,0 +1,130 @@
...
+* The layout component contains the Razor keyword `@Body`. During rendering, it's replaced by the content of the layout.
```
danroth27: > Razor keyward @blowdart 

It's not really a Razor keyword. How about: "The layout component uses the `Body` property to specify where the body content should be rendered. This is done using the Razor syntax `@Body`."

docs/layouts.md
```diff
@@ -0,0 +1,130 @@
...
+```csharp
+@layout MasterLayout
+
+@page "/master-data"
+
+<h2>Master Data Management</h2>
+...
+```
+
+## Centralized layout selection
```
danroth27: > The Blazor compiler includes the content of the view imports file

Blazor includes the directives specified in the view imports file

docs/layouts.md
guardrex: Yes, it should work out to something like (assuming `www`) ...

https://www.blazor.net/api/Microsoft.AspNetCore.Blazor.Layouts.ILayoutComponent.html

For reference, the current live link right now is ...

https://aspnet.github.io/Blazor.Docs/api/Microsoft.AspNetCore.Blazor.Layouts.ILayoutComponent.html

rstropek: @guardrex Should I use the *www.blazor.com* or the *aspnet.github.io" link? Should I turn all Blazor type references into API ref links?

guardrex: Use this one ...

```
https://www.blazor.net/api/Microsoft.AspNetCore.Blazor.Layouts.ILayoutComponent.html
```
... and I'll fix it later if it needs to change.

PR #43, opened 2018-04-16
Make it blue!

danroth27:
Proposing some styling updates:
- Blue gradient instead of brown for the hero
- Brighter green for the Get Started buttons and Experimental banner
- Use link styling in the swim lanes

image


danroth27:
:up::date: to fix layout issues (ex fix persistent horizontal scrollbar)

danroth27:
> The one thing I'd consider changing is the links. I feel like now they just blend into the text 

I prefer that that this links be less prominent since the main call to action on the home page is to Get Started.

guardrex:
It's cutting off the button a bit at the bottom ...

capture

guardrex:
Also, the B's don't match using the new font for "Blazor."

I'll also toss in that the new font for "Blazor" isn't particularly interesting imo. Perhaps, that's what you're going for, not too flashy. You could do what I did with the prior font/"Blazor" ... add just a smidgen of letter spacing ... just a few pixels ... to push the letters apart a bit.

capture

danroth27:
> Perhaps, that's what you're going for ... not too flashy.

Yup :smile:

PR #46, opened 2018-04-17
Update api refs to 0.2.0

danroth27:


guardrex:
Should `Microsoft.AspNetCore.Blazor.Server.dll` be removed? I tossed it in there on a 50-50 shot. (Now, you see why I don't vacation in Vegas! :smile:)

danroth27:
Just confirmed the dll list with Steve.

guardrex:
Note tho that the `Server` one is the RC one from myget.

PR #54, opened 2018-04-18
Open patch PRs on the master branch

guardrex:
I tested the patch process after setting up *master* and *live* branches, and it worked fine.

Now, we're getting patch PRs for the *live* branch. *Arrrrrrrgh!* :boy: :gun: :smile:

I have a question open with `@superyyrrzz` at DocFX to confirm this PR is the correct fix. Do you already know if this is correct? Alternatively, we could just roll the 🎲🎲 and see what happens!

### DIFF
--- a/docfx.json
+++ b/docfx.json
@@ -60,6 +60,14 @@
     "noLangKeyword": false,
     "keepFileLink": false,
     "cleanupCacheHistory": false,
-    "disableGitFeatures": false
+    "disableGitFeatures": false,
+    "globalMetadata": {
+      "newFileRepository": {
+        "branch": "master"
+      },
+      "_gitContribute": {
+        "branch": "master"
+      }
+    }
   }
 }


danroth27:
@vicancy could you please take a look at this as well? We want edit request to go against our master branch, not live. We only merge from master to live when we want to publish.

guardrex:
I asked over on DocFX, where another dev asked the same question. It was mostly clear what to do but not completely so.

https://github.com/dotnet/docfx/issues/482#issuecomment-382265734

superyyrrzz:
`newFileRepository` is obsolete. Set `_gitContribute` is enough in your case. Seems it's like the [same config](https://github.com/dotnet/docfx/blob/6715a94a2187f73ae3b1538afc6a13381f8a4114/Documentation/docfx.json#L47-L49) as DocFX's homepage. DocFX's document is built from master branch, whild `Improve This Doc` leads to dev branch.

guardrex:
Thanks @vicancy and @superyyrrzz ... working perfectly now. :+1:

PR #67, opened 2018-04-19
docfx.json updates to match aspnet/Docs

guardrex:
When I was looking at the problem with broken relative image links, I noticed a handful of DocFX config diffs between the config here and that of aspnet/Docs.

If these updates aren't sane ... or even just toooooo risky ... please just **close this**.

I don't want to cause an *international incident* here. :smile:

### DIFF
--- a/docfx.json
+++ b/docfx.json
@@ -17,38 +17,34 @@
       {
         "files": [
           "api/**.yml",
-          "api/index.md"
-        ]
-      },
-      {
-        "files": [
-          "docs/**.md",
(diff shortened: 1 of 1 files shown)


guardrex:
> I assume you have built locally and checked the site output that nothing weird got added or is missing.

Yes, I did build locally and didn't detect any problems. However, I want to do it *one more time* before I merge this. After merging, I'll immediately make live site checks.

guardrex:
@danroth27 I checked the production site ... everything seems good following this update. :+1:

PR #89, opened 2018-04-26
Refactor and enhance Child component language

### LINKED ISSUE #85: Child content section of the Components page should be updated
Hi,

it was founded a kind of typo in the **Child content** section of the **Components** page:

_ChildComponent.cshtml:_

```
<div class="panel panel-success">
  <div class="panel-heading">@Title</div>
  <div class="panel-body">@PanelContent</div>
</div>

@functions {
    public string Title { get; set; }
    public RenderFragment **PanelContent** { get; set; }
}
```
I got a fault when trying to execute this example. After a discussion in an issue report https://github.com/aspnet/Blazor/issues/686 it was founded the **PanelConent** field's name is incorrect. It is suggested to rename this to **ChildContent**. Unfortunatelly, there are still some errors with this name using. As a result I'm not sure how this example should be improved.

guardrex:
Fixes #85

### DIFF
--- a/docs/components/index.md
+++ b/docs/components/index.md
@@ -67,17 +67,20 @@ In the following example, the `ParentComponent` sets the value of the `Title` pr
 
 ## Child content
 
-Components can set the content of an element in another component by assigning to a property that represents a [RenderFragment](/api/Microsoft.AspNetCore.Blazor.RenderFragment.html) type.
-
-In the following example, the `ChildContent` element of the `ParentComponent` sets the value of the `ChildComponent`'s `ChildContent` property. `ChildContent` is a `RenderFragment` type and supplies the content to the Bootstrap panel.
+Components can set the content of an element in another component. The assigning component provides the content in the tags that specify the receiving component. For example, a `ParentComponent` can provide content that is to be rendered by a `ChildComponent` by placing the content inside **\<ChildComponent>** tags.
 
 *ParentComponent.cshtml*:
 
 [!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/ParentComponent.cshtml?start=1&end=7&highlight=6)]
 
+The receiving component has a `ChildContent` property that represents a [RenderFragment](/api/Microsoft.AspNetCore.Blazor.RenderFragment.html) type. The value of `ChildContent` is positioned in the markup where the content should be rendered.
+
 *ChildComponent.cshtml*:
 
 [!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/ChildComponent.cshtml?highlight=3,8)]
+
+> [!NOTE]
+> The property receiving the `RenderFragment` content must be named `ChildContent` by convention.
 
 ## Data binding
 
--- a/docs/components/index.md
+++ b/docs/components/index.md
@@ -67,13 +67,13 @@ In the following example, the `ParentComponent` sets the value of the `Title` pr
 
 ## Child content
 
-Components can set the content of an element in another component. The assigning component provides the content in the tags that specify the receiving component. For example, a `ParentComponent` can provide content that is to be rendered by a `ChildComponent` by placing the content inside **\<ChildComponent>** tags.
+Components can set the content in another component. The assigning component provides the content between the tags that specify the receiving component. For example, a `ParentComponent` can provide content that is to be rendered by a `ChildComponent` by placing the content inside **\<ChildComponent>** tags.
 
 *ParentComponent.cshtml*:
 
 [!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/ParentComponent.cshtml?start=1&end=7&highlight=6)]
 
-The receiving component has a `ChildContent` property that represents a [RenderFragment](/api/Microsoft.AspNetCore.Blazor.RenderFragment.html) type. The value of `ChildContent` is positioned in the markup where the content should be rendered.
+The child component has a `ChildContent` property that represents a [RenderFragment](/api/Microsoft.AspNetCore.Blazor.RenderFragment.html). The value of `ChildContent` is positioned in the child component's markup where the content should be rendered.
 
 *ChildComponent.cshtml*:
 
--- a/docs/components/index.md
+++ b/docs/components/index.md
@@ -73,7 +73,7 @@ Components can set the content in another component. The assigning component pro
 
 [!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/ParentComponent.cshtml?start=1&end=7&highlight=6)]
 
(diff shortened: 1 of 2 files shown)


docs/components/index.md
```diff
@@ -67,18 +67,21 @@ In the following example, the `ParentComponent` sets the value of the `Title` pr
 
 ## Child content
 
-Components can set the content of an element in another component by assigning to a property that represents a [RenderFragment](/api/Microsoft.AspNetCore.Blazor.RenderFragment.html) type.
-
-In the following example, the `ChildContent` element of the `ParentComponent` sets the value of the `ChildComponent`'s `ChildContent` property. `ChildContent` is a `RenderFragment` type and supplies the content to the Bootstrap panel.
+Components can set the content of an element in another component. The assigning component provides the content in the tags that specify the receiving component. For example, a `ParentComponent` can provide content that is to be rendered by a `ChildComponent` by placing the content inside **\<ChildComponent>** tags.
```
danroth27: This doesn't sound quite right to me. Child content allows a parent component to capture rendered content from a child component and then use it as part of its own rendering. I don't think the parent component provides content to the child content. I think it's the the other way around.

docs/components/index.md
```diff
@@ -67,18 +67,21 @@ In the following example, the `ParentComponent` sets the value of the `Title` pr
...
 
 *ParentComponent.cshtml*:
 
 [!code-cshtml[](common/samples/2.x/ComponentsSample/Pages/ParentComponent.cshtml?start=1&end=7&highlight=6)]
 
+The receiving component has a `ChildContent` property that represents a [RenderFragment](/api/Microsoft.AspNetCore.Blazor.RenderFragment.html) type. The value of `ChildContent` is positioned in the markup where the content should be rendered.
```
danroth27: The parent component captures child content by defining a property named `ChildContent` of type [RenderFragment](/api/Microsoft.AspNetCore.Blazor.RenderFragment.html). The parent component can then specify `@ChildContent` to render the child content as part of its own rendering.

docs/components/index.md
danroth27: High level that's pretty close. @SteveSandersonMS can speak better to the actual mechanics of how it works, but the high level the idea is that if you have a parent component with child content, like this:

```html
<Parent>
    // Child content with HTML, components, fancy Razor syntax, etc.
</Parent>
```

then the Parent component can capture the child content in a `RenderFragment` using a `ChildContent` property and then render it as part of the Parent component's rendered content.

docs/components/index.md
guardrex: We'll need to be very careful how we handle this.

> The parent component can then specify @ChildContent to render the child content as part of its own rendering.

The parent doesn't specify `@ChildContent` directly (I mean not in its markup by name, "ChildContent") ... it *appears* (just *looks like*) that the child does (in its property and its markup I mean).

If we try to explain it the way that you wrote it, I think we'll have a lot of head scratching.

Another phrase that concerns me ...

> parent component captures child content

... but "child content" can be confused: I don't want a reader to misconstrue that it's the entire rendered child in the parent.

Anyway ... back to my main point ... The "child" is the only place the dev will see an explicit reference to `ChildContent`/`@ChildContent`, so saying that the parent component specifies "ChildContent" is tricky to do without losing the reader. What I suggest is wording it more along the lines of the parent specifies the value of `ChildContent` in the value of the tags that specify the child component ... along those lines. I just want to make it really clear where `ChildContent` is and how it's set by the content of the tags in the parent.

docs/components/index.md
guardrex: I totally don't have it that way today ... and I find that somewhat confusing. The parent is a whole component. The child is a whole component. What I did was stick the child tags into the parent and then describe it based on how it's composed. Today, it goes like this ...

*ParentComponent.cshtml*:

```cshtml
@page "/ParentComponent"

<h1>Parent-child example</h1>

<ChildComponent Title="Panel Title from Parent">
    Panel content of the child component is supplied by the parent component.
</ChildComponent>
```

*ChildComponent.cshtml*:

```cshtml
<div class="panel panel-success">
  <div class="panel-heading">@Title</div>
  <div class="panel-body">@ChildContent</div>
</div>

@functions {
    public string Title { get; set; }
    public RenderFragment ChildContent { get; set; }
}
```

guardrex: I guess it boils down to how this is described and the sequence of what goes where and when ...

> have a parent component with child content

I didn't see it quite that way. The parent component only has the child content by virtue of the fact that the content is *inside of the tags that specify the child component.* If the child just looks up into the parent to pull that content down into its `RenderFragment` ... that's cool. Then, the child becomes fully rendered for the parent. I think the order is easy to state (if that order I wrote out is correct).

The greater challenge is making sure the reader doesn't get confused on what "child content" means. It's not the whole rendered child component. It's just the content inside the tags for the child component that end up in the `RenderFragment`. I'm assuming I'm correct about that. It was written and approved that way when we merged the topic.

I'm going to hold off on further changes until @SteveSandersonMS has a chance to look this over and say what's best.

guardrex:
My understanding is that **the fully rendered child component in the parent is not the "child content"** ... the "child content" is the content of the tags in the parent component that end up in the `ChildContent` `RenderFragment`. From that perspective, the text is correct. 🤔

@SteveSandersonMS Am I big 'ole *GooberRex*:tm:? 🙈 The problem might be just that the entire section is misnamed. This scenario might not be "Child content." If so, what's this called? I'll rename the section.

This is what the latest commit shows/says with the code inlined ...

> ## Child content

> Components can set the content in another component. The assigning component provides the element content between the tags that specify the receiving component. For example, a `ParentComponent` can provide content that is to be rendered by a `ChildComponent` by placing the content inside **\<ChildComponent>** tags.

> *ParentComponent.cshtml*:

```cshtml
@page "/ParentComponent"

<h1>Parent-child example</h1>

<ChildComponent Title="Panel Title from Parent">
    Child content of the child component is supplied by the parent component.
</ChildComponent>
```

> The child component has a `ChildContent` property that represents a [RenderFragment](/api/Microsoft.AspNetCore.Blazor.RenderFragment.html). The value of `ChildContent` is positioned in the child component's markup where the content should be rendered. In the following example, the value of `ChildContent` is received from the parent component and rendered inside the Bootstrap panel's `panel-body`.

*ChildComponent.cshtml*:

```cshtml
<div class="panel panel-success">
  <div class="panel-heading">@Title</div>
  <div class="panel-body">@ChildContent</div>
</div>

@functions {
    public string Title { get; set; }
    public RenderFragment ChildContent { get; set; }
}
```

> [!NOTE]
> The property receiving the `RenderFragment` content must be named `ChildContent` by convention.

PR #101, opened 2018-05-02
[WIP] Hosting static assets

### LINKED ISSUE #94: Hosting static assets
Note in *Host and deploy* pertaining to local hosting of static assets:

> Blazor sets up its own static file middleware to handle static file requests to content from the Blazor client app, not from the server project.

Note first tho that a CDN should be used for static assets and mention+link Azure Storage/CDN for this purpose.

Note that static assets for the client-side app can't be hosted in the server-side app using Static File Middleware at this time.

guardrex:
Fixes #94 

btw - That first paragraph just had a stray space at the end. I only deleted the space. You can skip that paragraph on review.


guardrex:
@danroth27 Same on this one. Just needs an :eye: on it to see if it's good.

StevenTCramer:
Other MS docs require each sentence to be on its own line.  MD will merge them together when generating html anyway.  This makes merges much easier.  For example your change of only removing a stray space.  Would be on 1 line vs showing and entire paragraph.

guardrex:
@StevenTCramer Our guidelines call for multi-line paragraphs in markdown. Yes indeed, it would have worked out just fine as you suggest to allow consecutive lines to roll into paragraphs. It would make it easier on us in two important ways ...

1. It would be easier to call out change requests on review because we wouldn't need to specify which sentence within a multi-line paragraph requires an adjustment.
1. As you say, we frequently blow GitHub's limit on how many characters they allow on a line in a diff to activate their change highlighting feature (red for drops; green for adds). It's so much of a pain that I even asked GH if they would consider increasing the limit. They said *No!* ... their attitude might change in the future. :wink:

There could be a downside tho. I don't think it adds a space between lines. I could be mistaken. When it happens these days that two consecutive lines roll together into a paragraph, it's by accident. We never do it on purpose, so I don't recall the exact behavior. If it's true that the needed space wouldn't be added automatically, then it would be left to us, and that would be a pain point of the consecutive line approach.

Well, that wasn't the direction they decided to go ... at least not in our wing of the docs (aspnet/Docs and dotnet/docs).

@Rick-Anderson Early on when MS moved docs to GH, was one sentence per line explicitly dismissed for some reason?

Rick-Anderson:
@StevenTCramer 
> Other MS docs require each sentence to be on its own line.

What docs?
Use rich diff - that works well in most cases.

@tdykstra  to comment. I don't think we want to use one sentence per line.

guardrex:
The problem I have with rich diff (thus the reason I don't use it) is that there is no intra-line highlighting at all. It makes it *even harder* for me to review PRs than source diff.

tdykstra:
>  I don't think we want to use one sentence per line.

I agree.  I have never seen anything like that in our contributor guide, which goes into a great deal of detail on Markdown style guidelines for all docs published on docs.microsoft.com.

StevenTCramer:
I am looking for the repo where I was told on my PR to change to one line.  But I can't find which repo it was just yet.  I thought an MS one, but maybe not.  I just published my Docs using DocFX and followed this convention and it works rather well.

I do not see a down side to it but the upside is much easier diffs.

StevenTCramer:
but if you google "[markdown sentence per line](https://www.google.co.th/search?rlz=1C1GCEA_enTH776TH777&ei=VIceW_CTO4jgrQH0uL24BA&q=markdown+sentence+per+line&oq=markdown+sentence+per+line&gs_l=psy-ab.3...5344.6127.0.6568.2.2.0.0.0.0.101.179.1j1.2.0....0...1.1.64.psy-ab..0.0.0....0.nxCCWBa7aBg)"  

You see that many others do this.

tdykstra:
@StevenTCramer Thanks for the follow-up.  I'll raise this question in our contributor guide working group to see if the team thinks we should consider moving in this direction.

guardrex:
@tdykstra I'm curious about the space between sentences, too. If no space is added automatically, it would be a major pain point imo.

Alternatively, it would be awesome to just have that line length cutoff changed that prevents us from seeing the change highlights on diffs. Idk the exact char count they use, but you know we blow that all the time with our paragraphs. If the acquisition goes thru, perhaps they'd be a bit more generous than when I inquired with them before about it.

tdykstra:
@guardrex I'll run a test to see if a space is added automatically.

tdykstra:
@guardrex The space is added automatically.

StevenTCramer:
Yes, sorry I forget to mention that the space is added automatically.  You can see it in my [blazor-state docs](https://timewarpengineering.github.io/blazor-state/) which are all done by breaking up the lines.

guardrex:
@tdykstra Very good. Thanks for testing. I still lean away from the proposal. Less scrolling is a good thing. 

My pain point is the review diff.

Either way, I just hope we get diff highlighting that will lead to 🏃 *FASTER* 🏃 reviewing.

docs/host-and-deploy/index.md
```diff
@@ -53,7 +53,13 @@ Because browsers make requests to Internet-based hosts for client-side pages, we
...
+
+## Static assets
+
+When a client-side Blazor app has access to the Internet, host large static assets (for example, images) on a Content Delivery Network (CDN), such as the [Azure CDN](https://azure.microsoft.com/services/cdn/). By hosting large static assets outside of the app, the app's payload size is reduced.
+
+Alternatively, a client-side Blazor app configures [Static File Middleware](https://docs.microsoft.com/aspnet/core/fundamentals/static-files) to handle static file requests within the app, not from the server project. Static assets can't be served from a Blazor server-side app to a Blazor client-side app outside of the app's initial load from the server. An attempt to add Static File Middleware to the server project with assets in the server project's *wwwroot* folder will fail to serve the assets to the client-side app.
```
danroth27: I'm sure what this is trying to say. I tried adding static file middleware and a wwwroot folder to my server project and the client Blazor app was able to use the static assets from the server project just fine.

guardrex: There may have been updates that made it possible after I tested. I was testing back before May 1, so perhaps things have changed. I'll try again ... *as soon as I can get free.* 😅 🏃

guardrex:
@danroth27 Closing this ... works fine. I think it was just a point-in-time problem.

PR #113, opened 2018-05-04
Fixing build errors in 0.3

floreseken:
This current example code does not compile in 0.3, I fixed that. But the text would probably need some changes too..

### DIFF
--- a/docs/components/index.md
+++ b/docs/components/index.md
@@ -138,10 +138,10 @@ Child component:
 
 @functions {
     [Parameter]
-    public int Year { get; set; }
+    private int Year { get; set; }
     
     [Parameter]
-    public Action<int> YearChanged { get; set; }
+    private Action<int> YearChanged { get; set; }
 }
 ```
 
--- a/docs/components/index.md
+++ b/docs/components/index.md
@@ -111,7 +111,8 @@ Data binding works with [DateTime](https://docs.microsoft.com/dotnet/api/system.
 <input bind="@StartDate" format-value="yyyy-MM-dd" />
 
 @functions {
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @floreseken!

@danroth27 A handful of `public` props seem like they would change to `private` and get the attribute ...

Line 114: `public DateTime StartDate`

Line 130: `public int ParentYear`

Line 374: `public bool IsCompleted`

floreseken:
Yes that too, forgot about the private part..

guardrex:
When Dan confirms that those others should be changed (I think so since we marked them `public` and thus implied that they would be settable from an external component), then one of us can square them all away here.

docs/components/index.md
```diff
@@ -137,7 +137,10 @@ Child component:
 <div> ... </div>
 
 @functions {
+    [Parameter]
     public int Year { get; set; }
```
danroth27: Should be private

docs/components/index.md
```diff
@@ -137,7 +137,10 @@ Child component:
...
+    
+    [Parameter]
```
danroth27: @rynowak Is `[Parameter]` needed here?

guardrex:
... and these others ... `[Parameter]` + `private`, too?

Line 114: public DateTime StartDate
Line 130: public int ParentYear
Line 374: public bool IsCompleted

guardrex:
@floreseken They wired this repo to require the CLA signature. Would you mind reviewing and signing the CLA?

@danroth27 I made the other changes. This LGTM ... u ok with these updates?

PR #116, opened 2018-05-06
Add BlazorStrap

iAmBipinPaul:
Bootstrap 4 Components for Blazor Framework

### DIFF
--- a/community.md
+++ b/community.md
@@ -19,13 +19,11 @@ Below is a list of projects and resources from the Blazor community. If you know
 ## Tooling
 * [Razor+ Visual Studio Code extension](https://marketplace.visualstudio.com/items?itemName=austincummings.razor-plus)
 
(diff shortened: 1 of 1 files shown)


community.md
```diff
@@ -24,5 +24,8 @@ Below is a list of projects and resources from the Blazor community. If you know
 * [Blazor local and session storage support](https://github.com/cloudcrate/BlazorStorage)
 * [BlazorComponents (ChartJS Interop)](https://github.com/muqeet-khan/BlazorComponents)
 
+## Bootstrap components
```
danroth27: How about we rename the previous section to just "Libraries" and move BlazorComponents into that section?

iAmBipinPaul: Yeah , that will be nice to have all interop under one section. 👍

PR #118, opened 2018-05-08
I guess it is typo error

babula38:
I guess this is a typo error, i created a pull request against master as i have very less knowledge what to do.

if you have any guide just share with me, that would be helpfull


guardrex:
Thanks @babula38 ... It was taken care of by the author @rstropek on https://github.com/aspnet/Blazor.Docs/pull/119.

PR #119, opened 2018-05-08
Fix incorrect constructor name

### LINKED ISSUE #112: Class name and constructor name does not match
In the [Dependency injection in services](https://blazor.net/docs/dependency-injection.html#dependency-injection-in-services) section,
the class and the constructor don’t match

public class **DataAccess** : IDataAccess
{
    // The constructor receives an HttpClient via dependency
    // injection. HttpClient is a default service offered by Blazor.
    public **Repository**(HttpClient client)
    {
        ...
    }
    ...
}

rstropek:
Fixes #112


guardrex:
Thanks @rstropek!

PR #124, opened 2018-05-14
Added link to Blazor-Fluxor Redux/Flux library for Blazor

mrpmorris:


mrpmorris:
I added the link in alphabetical order, I thought that would prevent future race-conditions for the top spot. Perhaps that strategy should be employed for all links?

guardrex:
> Perhaps that strategy should be employed for all links?

I agree. :+1:

PR #165, opened 2018-06-26
[WIP] Added ref to Toss

RemiBou:
This project uses Blazor for client with an asp.net core api, cosmosdb and the aspnet core security api


RemiBou:
No problem, I'll tell you here when it's done

RemiBou:
https://github.com/RemiBou/Toss.Blazor/blob/master/README.md

PR #177, opened 2018-07-16
Add note to mention adding event listeners that need to call C# methods

chassq:
Added this to learn blazor docs. See https://learn-blazor.com/architecture/interop/ . So thought about adding it here as well.

### DIFF
--- a/docs/javascript-interop.md
+++ b/docs/javascript-interop.md
@@ -111,6 +111,8 @@ To call a .NET method, create the method in C# and invoke the method from JavaSc
 
 ## Share interop code in a Blazor class library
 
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks for the submission @chassq. I'll make some small changes to the text to conform to MS guidelines after we hear back from engineering.

guardrex:
@chassq Do you want to update the PR for 0.5.0?

guardrex:
@danroth27 Should this be closed, or should I work the concepts in the PR forward for the latest JS interop bits?

danroth27:
I think we can just close this one.

PR #189, opened 2018-07-25
Add blazor hosting models topic

danroth27:
Rename "Host and deploy" to "Publishing" and add new "Hosting models" doc to cover client-side vs server-side Blazor.

### DIFF
--- a/docs/host-and-deploy/hosting-models.md
+++ b/docs/host-and-deploy/hosting-models.md
@@ -9,7 +9,7 @@ ms.date: 07/24/2018
 ms.prod: asp.net-core
 ms.technology: aspnet
 ms.topic: article
-uid: client-side/blazor/publish-and-deploy/hosting-models
+uid: client-side/blazor/host-and-deploy/hosting-models
 ---
 # Blazor hosting models
 
@@ -21,7 +21,7 @@ Blazor is a client-side web framework designed primarily to run in the browser o
 
 ## Client-side (in-process) hosting model
 
-The principal hosting model for Blazor is running client-side in the browser. In this model, the Blazor app, its dependencies, and the .NET runtime are downloaded to the browser, and the app is executed directly on the browser UI thread. All UI updates and event handling happens within the same process. The app assets can be published as static files using whatever web server is preferred (see [Publish and deploy](xref:client-side/blazor/publish-and-deploy/index)).
+The principal hosting model for Blazor is running client-side in the browser. In this model, the Blazor app, its dependencies, and the .NET runtime are downloaded to the browser, and the app is executed directly on the browser UI thread. All UI updates and event handling happens within the same process. The app assets can be deployed as static files using whatever web server is preferred (see [Host and deploy](xref:client-side/blazor/host-and-deploy/index)).
 
 ![Blazor client-side](https://user-images.githubusercontent.com/1874516/43042852-998bb680-8d3b-11e8-9d39-adf8d3d77360.png)
 
--- a/docs/host-and-deploy/hosting-models.md
+++ b/docs/host-and-deploy/hosting-models.md
@@ -9,7 +9,7 @@ ms.date: 07/24/2018
 ms.prod: asp.net-core
 ms.technology: aspnet
 ms.topic: article
-uid: client-side/blazor/hosting-models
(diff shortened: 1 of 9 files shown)


guardrex:
RE: The TOC change (*Host and deploy* to *Publishing*) ... There's a minor concern and a more serious APEX guidance issue.

The minor concern first: We went the other way with ASP.NET Core docs. We went from *Publishing* to *Host and Deploy* partially because the "publishing" section of TOC held topics and topic content that also pertained to server hosting, which we surmised isn't the same as "publishing." I think we also felt that "deployments" are more inline with IT industry activities than "publishing," which feels more print industry related. These concerns will be mirrored in the Blazor section of the TOC, where content will also pertain to hosting/serving Blazor apps.

The more serious issue: APEX has issued guidance that no gerunds should be used in titles or section headings of topics unless necessary on the grounds that they don't localize well. Therefore, we were left with having to use the word "publish" over "publishing." When it came to consideration of TOC titles such as *Publish and deploy* or *Host and deploy*, the team (and Rick's final choice) favored *Host and deploy*. If you prefer the change, APEX approves of "publish" over "publishing."

Anyway, I'm sure they won't call out the firing squad on me :boy::gun: if you want to go with *Publishing*. :smile:

guardrex:
Sounds good. I'll leave this alone until u give the word.

danroth27:
I changed my mind again 😝 

Hosting is Blazor is subtly different than hosting in ASP.NET Core. Hosting Blazor refers to where the Blazor rendering logic is running (i.e. in the browser UI thread, in a web worker, in Electron, on the server, etc.). Hosting in ASP.NET Core refers to hosting the app. So I decided to go with two separate docs. I removed stuff from the Blazor hosting models doc that is covered in the deployment models section of the renamed Publish and deploy doc.

guardrex:
I see what you mean, but it feels confusing taking these terms at face value. "Publish" and "deploy" are generally synonymous terms. The words themselves don't indicate *where* each app is located, what's being *deployed* where, and which app *serves* what.

If this is all part of the ordinary SPA experience, it's not surprising that I would be confused by it. I'm a web dev. :smile: ... learning more about SPA approaches where/when I can.

Ultimately the problem is that when it comes to Blazor (based on your remarks above), there are **_two_** "hosts" and there are **_two_** "deployments." That means that using the terms will be instantly confusing to anyone coming from a web dev background into this for the first time.

I still recommend *Host and deploy*. Given it's just a title and doesn't say what's *hosted* where or what's *deployed* where, it works. "Host" refers to the act of an app being somewhere (server or client; on a dev box for debugging or on a server ... or a client ... where one or more apps sit and interact with users or other apps); "deploy" refers to the act of moving one or more apps somewhere (server or client; from a dev environment to a server or from a server to a client).

danroth27:
> I still recommend Host and deploy

And then next the hosting models topic underneath?

guardrex:
Yes, or the reverse if you think that that's better introductory info. What you wrote seemed like great first contact content in this area ... explaining the basics/general setup before details are discussed. I've looked back at what I did before in https://blazor.net/docs/host-and-deploy/index.html#deployment-models, and I can see that it could've gone the other way ... That *Deployment models* content could have been only the landing page content with the rest on a subtopic or at least moved to the top of the current topic.

The overall goal tho is to mirror the layout in the aspnet/Docs (although this will all end up under *client-side* later) ...

Today:

* Host and deploy (landing page; index) - Everything there, since there wasn't much content at the time we started.

To:

* Host and deploy (landing page; index) - Basic information and links to subtopics pertaining to host and deployment scenarios
  - Subtopic 1 under the index for specific scenario content
  - Subtopic 2 ... another scenario
  - So on ...

guardrex:
It's borne out in the high-level planning for this part of the TOC. Scroll down to the *Host and deploy* area of ... https://github.com/aspnet/Blazor.Docs/wiki/Blazor-high-level-planning ... and you'll see what I mean. Those subtopics can get into all of the juicy details on each scenario. The landing page is more of a basic overview (along the lines of what's in your new content) and a jumping off point for the subtopics.

danroth27:
@guardrex Done

guardrex:
@danroth27 Cool. Do u still want to wait on the other reviewers before merging this to master?

PR #190, opened 2018-07-25
Add Blazor debugging doc

danroth27:


guardrex:
I'll perform a quick guidelines pass ... just the critical APEX updates to make them happy. I'll add a review after that because there are a couple of optional items that you may or may not want to address.

docs/debugging.md
```diff
@@ -0,0 +1,58 @@
+---
+title: Blazor debugging
```
guardrex: The gerund thing again ... APEX would prefer "Debug Blazor" or something that avoids a gerund here.

docs/debugging.md
```diff
@@ -0,0 +1,58 @@
...
+author: danroth27
+description: Learn how to debug Blazor apps.
+manager: wpickett
+ms.author: riande
+ms.custom: mvc
+ms.date: 07/25/2018
+ms.prod: asp.net-core
+ms.technology: aspnet
+ms.topic: article
+uid: client-side/blazor/debugging
```
guardrex: Same here with what is also going to be the URL segment. They would prefer `client-side/blazor/debug`.

docs/debugging.md
```diff
@@ -0,0 +1,58 @@
...
+---
+# Blazor debugging
+
+[Daniel Roth](https://github.com/danroth27)
+
+[!INCLUDE[](~/includes/blazor-preview-notice.md)]
+
+Blazor has some *very early* support for debugging client-side Blazor apps running on WebAssembly in Chrome. While this initial debugging support is very limited and unpolished, it shows the basic debugging infrastructure coming together.
```
guardrex: I didn't edit these "very" usages out, but they've fallen way out of favor in technical writing. They don't convey any useful information because they're an unfathomable degree of emphasis.

WRT APEX, they don't explicitly prevent the use of the word; however, they do push the "be concise" rule quite a bit. Therefore, they would probably be happy if the word "very" saw limited use.

If you want me to nix them, I'll factor them out. They really aren't *very* :smile: necessary. lol

danroth27:
@guardrex These updates look fine. Merge it?

docs/debugging.md
```diff
@@ -0,0 +1,58 @@
...
+manager: wpickett
+ms.author: riande
+ms.custom: mvc
+ms.date: 07/25/2018
+ms.prod: asp.net-core
+ms.technology: aspnet
+ms.topic: article
+uid: client-side/blazor/debugging
+---
+# Blazor debugging
```
guardrex: ... and here.

guardrex:
@danroth27 Sure ... go ahead if you're ok with the review notes.

docs/debugging.md
guardrex: Same deal for the name of the file ... *debug.md*.

danroth27:
I removed one "very" but I'd like to leave the "very limited" phrase as I think it's important to emphasize that only few debugging scenarios currently work.

guardrex:
> I removed one "very" but I'd like to leave the "very limited" phrase as I think it's important to emphasize that only few debugging scenarios currently work.

That's cool. APEX is less sensitive on this one. It's more of a technical writing standard, and the Chicago manual *may* say something about it. It's been pushed in technical writing circles for years now. Writers don't like *very* very much. lol ... just not precise enough for their tastes.

PR #197, opened 2018-07-26
Update tutorial sample to 0.5.1

### LINKED ISSUE #196: Update tutorial sample to 0.5.0

guardrex:
Fixes #196 

The `blazor` template doesn't have the weather data paging, so I'm removing it here. It was never part of the tutorial to explain it (and when the reader `dotnet new`s-up the app, it won't have it), so it can be cut. If you prefer that I re-introduce that to the sample, let me know.


guardrex:
@danroth27 Should we hold off here until 0.5.1 lands, or should we go ahead here and then do a quick 0.5.1 update to the samples later?

danroth27:
@guardrex  We expect to release the patch this morning, so I think we should just wait.

danroth27:
Blazor 0.5.1 is now live

guardrex:
Thanks @danroth27 ... I'll get both of these PRs fixed up this evening (Friday) and ping u when they're ready to go.

guardrex:
@danroth27 This one is ready as well.

PR #202, opened 2018-07-30
Added BlazeDown to Samples

EdCharbeneau:
Added BlazeDown, an online Markdown editor built with Blazor, to the samples list.


danroth27:
@EdCharbeneau are you going to update the demo with the 0.5.1 version? it looks like the demo is still using older bits.

EdCharbeneau:
@danroth27 Yes. The source code has been updated. The hosted demo hasn't gotten the new bits yet.

PR #229, opened 2018-08-19
Document Blazor dev server --pathbase arg

### LINKED ISSUE #154: Document Blazor dev server parameters
Ex `--pathbase`

guardrex:
Fixes #154 

Is `--pathbase` the only one?


danroth27:
> Is --pathbase the only one?

Technically I believe you can pass arguments corresponding to any of the normal ASP.NET Core host configuration values: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/web-host#host-configuration-values

guardrex:
🤔 I think we should go with a section to describe it. A `[!NOTE]` in the *App base path* section seems out of place. Such a section will probably need to appear *after* the deployment models are discussed. I'll try that on the next commit, and we can see how it looks.

guardrex:
@danroth27 I tried a test of `--webroot=public` with a *public* folder and an Index component image tag requesting the image file. No dice ... it still went for *wwwroot*.

danroth27:
Hmm . . . it looks like `--urls=<url>` works. @Tratcher what support do we have for configuring hosting via command-line args?

Tratcher:
@danroth27 it should work so long as you pass the command line args to CreateDefaultWebHostBuilder or similar.

guardrex:
@Tratcher Any idea why my test with `--webroot=public` and the env var (`"ASPNETCORE_WEBROOT": "public"`) didn't work? The setup was fairly straightforward ...

* Server project with a *public* folder and an image inside it. Also, a different image sitting in *wwwroot*.
* Started the project with the `--webroot=public` switch. (I also tried the env var set in *launchSettings.json*).
* Blazor component with an `` tag that just asks for the image sitting in *public* (no path or anything ... just the file name).
* A second `` tag asking for the image sitting in *wwwroot* in the server project.

... the *wwwroot* image is served. The image in the *public* folder isn't served.

Tratcher:
Have the project code available somewhere?

guardrex:
https://github.com/guardrex/BlazorApp1

It's pretty much a template project.

The Index component has the two `` tags.

One image is in *wwwroot*. The other is in *public*.

Wanted to try `UseWebRoot`, but it isn't available ...

> 'IWebAssemblyHostBuilder' does not contain a definition for 'UseWebRoot' and no extension method 'UseWebRoot' accepting a first argument of type 'IWebAssemblyHostBuilder' could be found (are you missing a using directive or an assembly reference?)

Tratcher:
Well you're dropping the command line args to begin with:
https://github.com/guardrex/BlazorApp1/blob/3a70e611fe34ecdb0ca3450ee9c660bf21e611a5/Program.cs#L13

guardrex:
Hum ...... I wonder if I did that or the template came that way. I'll set another one up.

btw tho ... setting the env var didn't work either.

Tratcher:
Ah, this is the first time I've really looked into the blazor code, It's not even using webhost. There is no config.
https://github.com/aspnet/Blazor/blob/master/src/Microsoft.AspNetCore.Blazor.Browser/Hosting/WebAssemblyHostBuilder.cs

guardrex:
I see.

So it looks like this is the answer to my original question ...

https://github.com/aspnet/Blazor/blob/f72e03f8fe46735436f80950cac49808fc3e3c11/test/Microsoft.AspNetCore.Blazor.E2ETest/Infrastructure/ServerFixtures/DevHostServerFixture.cs

guardrex:
I'll add `--urls` and `--contentroot`. @danroth27 ... I'll mark this WIP for now and get back to this on Sunday.

I'm 🏃🏃🏃🏃🏃🏃😅 with aspnet/Docs issues at the moment.

guardrex:
Side-note ... I just created a new project from the template, and it gave me ...

```csharp
namespace BlazorApp4
{
    public class Program
    {
        public static void Main(string[] args)
        {
            CreateHostBuilder(args).Build().Run();
        }

        public static IWebAssemblyHostBuilder CreateHostBuilder(string[] args) =>
            BlazorWebAssemblyHost.CreateDefaultBuilder()
                .UseBlazorStartup<Startup>();
    }
}
```

I assume that it isn't a mistake since args do get picked up (e.g., @danroth27's test with `--urls`).

danroth27:
@Tratcher I believe the relevant code here is actually in the Blazor dev server: https://github.com/aspnet/Blazor/tree/master/src/Microsoft.AspNetCore.Blazor.Cli

Tratcher:
Ah, here's the problem. The blazor dev server doesn't have a normal static file middleware. Blazor sets up static files without webroot, it uses its own config:
https://github.com/aspnet/Blazor/blob/39d2b90f03b037cf4601a056b37ab55fad9daab6/src/Microsoft.AspNetCore.Blazor.Server/Builder/BlazorApplicationBuilderExtensions.cs#L66-L71

guardrex:
@danroth27 

The last commit includes the following updates:

* A touch-up for rendering tags. aspnet/Docs break with the style manual now and code-fence markup tags, so I do that here in anticipation of these topics joining the aspnet/Docs one day.
* I include some content (IMPORTANT notes) for removing the ANCM handler if a client-side Blazor app is hosted as a sub-app under an ASP.NET Core app in IIS. I also make it clear to the reader that this is in addition to following the app base path guidance and setting the `<base>` tag to the IIS alias for the sub-app. I also explicitly state that this applies to Blazor apps based on the **Blazor** project template.
* The main thing here are the host configuration values: Apparently, server-side hosting model apps (based on an ASP.NET Core app) can use any of the host config values in the Web Host topic over in aspnet/Docs, so I link over for that scenario. For the two client-side scenarios, there are `--pathbase`, `--contentroot`, and `--urls`. Those are explained out in the new section on host configuration values with explicit examples on how to set them. I use the same example, `CoolBlazorApp`, throughout. I make it clear when the trailing slash is required.

*Apparently == Some of this content is based on testing here, and I have a degree of confidence in those areas. Some of this tho is based on *speculation*. Given my aspnet/Docs workload, I don't have time to read a lot of source code, so please excuse me where I go off the rails.

**Side-note:** Having to deal with "app base path" and `--pathbase` (i.e., "base path" vs. "path base") seems unfortunate. Is there merit to my suggesting that the `--pathbase` switch be recast as `--basepath` (or `--appbasepath`) to match? ... I mean before it's too late ... and by "too late," I mean after Blazor becomes a product and we're stuck with the switch name forever.

guardrex:
@danroth27 This one is ready for a final look/changes.

guardrex:
I resolved the merge conflict here. Anything else to do on this one?

guardrex:
@danroth27 This should be ready, too. All good?

docs/host-and-deploy/hosting-models.md
```diff
@@ -45,13 +45,27 @@ The downsides to the client-side hosting model are:
 * Larger download size and app load time
 * Less mature .NET runtime and tooling support (for example, limitations in .NET Standard support and debugging)
 
+> [!NOTE]
+> Visual Studio includes the **Blazor (ASP.NET Core hosted)** project template for creating a Blazor app that runs on WebAssembly and is hosted on an ASP.NET Core server. The server-side app merely serves the Blazor app to clients. The server-side app doesn't directly interact with the app after serving it to clients other than to possibly serve static assets from its web root (for example, *wwwroot*).
```
danroth27: server-side app -> ASP.NET Core app

danroth27: > The server-side app doesn't directly interact with the app after serving it to clients other than to possibly serve static assets from its web root (for example, *wwwroot*). 

You can have other interactions. For example, you might use a Web API or a SignalR Hub. 

How about something like this?:

Visual Studio includes the **Blazor (ASP.NET Core hosted)** project template for creating a Blazor app that runs on WebAssembly and is hosted on an ASP.NET Core server. The ASP.NET Core app handles serves the Blazor app to clients, but is otherwise a separate process. The client-side Blazor app can interact with the server over the network using web API calls or SignalR connections.

guardrex:
@danroth27 Changes made. :+1:

docs/host-and-deploy/hosting-models.md
```diff
@@ -45,13 +45,27 @@ The downsides to the client-side hosting model are:
 * Larger download size and app load time
 * Less mature .NET runtime and tooling support (for example, limitations in .NET Standard support and debugging)
 
+> [!NOTE]
+> Visual Studio includes the **Blazor (ASP.NET Core hosted)** project template for creating a Blazor app that runs on WebAssembly and is hosted on an ASP.NET Core server. The ASP.NET Core app serves the Blazor app to clients but is otherwise a separate process. The client-side Blazor app can interact with the server over the network using Web API calls or SignalR connections.
+
+> [!IMPORTANT]
+> If a Blazor app (based on the **Blazor** project template) uses the client-side (in-process) hosting model and is hosted as an IIS sub-application in an ASP.NET Core app, it's important to disable the inherited ASP.NET Core Module handler. Remove the handler in the Blazor app's published *web.config* file by adding a `<handlers>` section to the file:
```
danroth27: > If a Blazor app (based on the **Blazor** project template) uses the client-side (in-process) hosting model and is hosted as an IIS sub-application in an ASP.NET Core app

I had to read this a bunch of times to figure out what it was saying. Is there a way to simplify this text? How about:

If a client-side Blazor app is hosted in an ASP.NET Core app that is hosted in an IIS sub-application, ...

danroth27: I don't remember why we decided to add "(in-process)" when referring to the client-side hosting model, but I think it's unnecessary and confusion. Can we scrub "(in-process)" from this doc?

docs/host-and-deploy/index.md
```diff
@@ -55,9 +55,101 @@ Because browsers make requests to Internet-based hosts for client-side pages, we
...
+```console
+dotnet run --pathbase=/CoolBlazorApp
+```
+
+The app responds locally at `http://localhost:port/CoolBlazorApp`.
+
+For more information, see the [path base host configuration value](#path-base) section.
+
+> [!IMPORTANT]
+> If a Blazor app uses the [client-side (in-process) hosting model](xref:client-side/blazor/host-and-deploy/hosting-models#client-side-in-process-hosting-model) (based on the **Blazor** project template) and is hosted as an IIS sub-application in an ASP.NET Core app, it's important to disable the inherited ASP.NET Core Module handler. Remove the handler in the Blazor app's published *web.config* file by adding a `<handlers>` section to the file:
```
danroth27: Same comments as above.

docs/host-and-deploy/index.md
```diff
@@ -55,9 +55,101 @@ Because browsers make requests to Internet-based hosts for client-side pages, we
...
+> </handlers>
+> ```
+>
+> Removing the handler is performed in addition to configuring the app's base path as described in this section. Set the app base path in the Blazor app's *index.html* file to the IIS alias used when configuring the sub-app in IIS.
+
+## Host configuration values
+
+Blazor apps that use the [server-side hosting model](xref:client-side/blazor/host-and-deploy/hosting-models#server-side-hosting-model) can accept [Web Host configuration values](https://docs.microsoft.com/aspnet/core/fundamentals/host/web-host#host-configuration-values).
+
+Blazor apps that use the [client-side (in-process) hosting model](xref:client-side/blazor/host-and-deploy/hosting-models#client-side-in-process-hosting-model) can accept the following host configuration values as command-line arguments at runtime.
```
danroth27: I believe these are really **dev** host specific settings, right? Once you publish the app it's just a bunch of static files and there is no command-line provided host configuration values involved.

docs/host-and-deploy/index.md
```diff
@@ -55,9 +55,101 @@ Because browsers make requests to Internet-based hosts for client-side pages, we
...
+
+### Content Root
+
+The `--contentroot` argument determines where the app begins searching for content files, such as components.
```
danroth27: The app doesn't search for components using the file system. Components get compiled into classes that are then discovered as such at runtime. The content root is "the base path to any content used by the app". For example the web root is typically relative to the content root.

guardrex:
@danroth27 :+1:

PR #230, opened 2018-08-19
Document Route parameters with type

### LINKED ISSUE #183: Document Route parameters with type
Existing Topic: https://blazor.net/docs/routing.html

Currenty it's possible to cast the parameter directly to the property type by adding `:type`. Currently it's not yet supported in the docs.
 
**An example code snippet:**

```csharp
@page "/Users"
@page "/Users/{Id:int}"

<h1>The user Id is @Id !</h1>

@functions {
    [Parameter]    private int Id { get; set; }
}
```

guardrex:
Fixes #183 

This assumes that all of the ASP.NET Core constraints are applicable.

cc: @vertonghenb

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -67,7 +67,21 @@ In the following example, the route to the Users component only matches if:
 }
 ```
 
-For more information and a list of available route constraints, see [ASP.NET Core Routing: Route constraint reference](https://docs.microsoft.com/aspnet/core/fundamentals/routing#route-constraint-reference).
+The route constraints shown the following table are available for use. For the route constraints that match with the invariant culture, see the warning below the table for more information.
+
+| Constraint | Example           | Example Matches                        | Invariant<br>culture<br>matching | Notes |
+| ---------- | ----------------- | -------------------------------------- | :------------------------------: | ----- |
+| `bool`     | `{active:bool}`   | `true`, `FALSE`                        | No                               | Matches `true` or `false` (case-insensitive) |
+| `datetime` | `{dob:datetime}`  | `2016-12-31`, `2016-12-31 7:32pm`      | Yes                              | Matches a valid `DateTime` value |
+| `decimal`  | `{price:decimal}` | `49.99`, `-1,000.01`                   | Yes                              | Matches a valid `decimal` value |
+| `double`   | `{weight:double}` | `1.234`, `-1,001.01e8`                 | Yes                              | Matches a valid `double` value |
+| `float`    | `{weight:float}`  | `1.234`, `-1,001.01e8`                 | Yes                              | Matches a valid `float` value |
+| `guid`     | `{id:guid}`       | `CD2C1638-1638-72D5-1638-DEADBEEF1638` | No                               | Matches a valid `Guid` value, including when in braces (`{<GUID>}`) |
+| `int`      | `{id:int}`        | `123456789`, `-123456789`              | Yes                              | Matches a valid `int` |
+| `long`     | `{ticks:long}`    | `123456789`, `-123456789`              | Yes                              | Matches a valid `long` value |
+
+> [!WARNING]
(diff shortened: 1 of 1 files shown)


danroth27:
Here's what we currently support in Blazor: https://github.com/aspnet/Blazor/blob/master/src/Microsoft.AspNetCore.Blazor/Routing/RouteConstraint.cs#L43-L71

guardrex:
@danroth27 Added a table of the supported constraints.

It appears that the warning applies ... the methods in aspnet/Routing constraints are similarly constructed.

docs/routing.md
```diff
@@ -47,6 +47,42 @@ The Blazor client-side router uses route parameters to populate the correspondin
...
+| `float`    | `{weight:float}`  | `1.234`, `-1,001.01e8`                                                           | Yes                              |
+| `guid`     | `{id:guid}`       | `CD2C1638-1638-72D5-1638-DEADBEEF1638`, `{CD2C1638-1638-72D5-1638-DEADBEEF1638}` | No                               |
+| `int`      | `{id:int}`        | `123456789`, `-123456789`                                                        | Yes                              |
+| `long`     | `{ticks:long}`    | `123456789`, `-123456789`                                                        | Yes                              |
+
+> [!WARNING]
+> Route constraints that verify the URL and are converted to a CLR type (such as `int` or `DateTime`) always use the invariant culture. These constraints assume that the URL is non-localizable. The framework-provided route constraints don't modify the values stored in route values. All route values parsed from the URL are stored as strings. For example, the `float` constraint attempts to convert the route value to a float, but the converted value is used only to verify that it can be converted to a float.
```
danroth27: I don't know that the second part of this applies (starting with the 3rd sentence). @rynowak Can you take a look?

docs/routing.md
```diff
@@ -47,6 +47,42 @@ The Blazor client-side router uses route parameters to populate the correspondin
...
+
+<h1>The user Id is @Id!</h1>
+
+@functions {
+    [Parameter]
+    private int Id { get; set; }
+}
+```
+
+The route constraints shown the following table are available for use. For the route constraints that match with the invariant culture, see the warning below the table for more information.
```
danroth27: shown in the

guardrex:
ping @rynowak

docs/routing.md
guardrex: @rynowak 

We need to confirm if the 2nd warning content in our ASP.NET Core docs on route constraints here ...

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-2.1#route-constraint-reference

... also applies to route constraints for Blazor (I brought that text over to Line 84 here) ...

> The framework-provided route constraints don't modify the values stored in route values. All route values parsed from the URL are stored as strings. For example, the `float` constraint attempts to convert the route value to a float, but the converted value is used only to verify that it can be converted to a float.

The code is over here :point_right: https://github.com/aspnet/Blazor/blob/master/src/Microsoft.AspNetCore.Blazor/Routing/RouteConstraint.cs#L43-L71

guardrex:
@rynowak This is a quick ❓ ...

We need to confirm if the 2nd warning content in our ASP.NET Core docs on route constraints here ...

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-2.1#route-constraint-reference

... also applies to route constraints for Blazor (I brought that text over to Line 84 here) ...

> The framework-provided route constraints don't modify the values stored in route values. All route values parsed from the URL are stored as strings. For example, the `float` constraint attempts to convert the route value to a float, but the converted value is used only to verify that it can be converted to a float.

The code is over here :point_right: https://github.com/aspnet/Blazor/blob/master/src/Microsoft.AspNetCore.Blazor/Routing/RouteConstraint.cs#L43-L71

rynowak:
It looks like Blazor's route constraints **DO** convert values

guardrex:
Thanks @rynowak ... I'll update the PR in a few days when I get back from vacation. 🏖:beer::smile:

guardrex:
@danroth27 I've updated this based on @rynowak feedback.

PR #235, opened 2018-09-02
Add JS interop to sample app

### LINKED ISSUE #150: Add JS interop example to sample app
Blazor JavaScript and TypeScript interop
docs/javascript-typescript-interop.md

Add the JS interop example into the sample app.

### LINKED ISSUE #257: Wording in Interop help
Blazor docs on page https://blazor.net/docs/javascript-interop.html has following text.

```
The IJSRuntime abstraction is asynchronous to allow for out-of-process scenarios.
If the app runs in-process and you want to invoke a JavaScript function synchronously, 
downcast to IJSInProcessRuntime and call Invoke<T> instead. We recommend that 
most JavaScript interop libraries use the async APIs to ensure the libraries are available 
in all Blazor scenarios, client-side or server-side.
```

I'm a confused by the term in-process and out-of-process scenarios. It is not clear is client-side code considered in-process , or out-of-process by the team. Would be good to clarify that point.

guardrex:
Fixes #150
Fixes #257


guardrex:
@danroth27 Try it now. :eyes:

The problem that I had before was just calling a method wrong ... there was perhaps one-too-many "SayHello" things in the mix. :smile:

I've attempted to avoid that confusion by naming the method that calls `HelloHelper.SayHello` `ExampleJsInterop.CallHelloHelperSayHello` rather than just `ExampleJsInterop.SayHello`, which is where I went wrong when I was trying to get this to work the first time around.

docs/javascript-interop.md
```diff
@@ -23,35 +23,49 @@ A Blazor app can invoke JavaScript functions from .NET and .NET methods from Jav
...
-    }
-}
-```
+[!code-javascript[](./common/samples/2.x/BlazorSample/wwwroot/exampleJsInterop.js?highlight=2-7)]
+
+Place the `<script>` tag that references the JavaScript file in the *wwwroot/index.html* file:
+
+[!code-html[](./common/samples/2.x/BlazorSample/wwwroot/index.html?highlight=16)]
+
+Don't place a script tag in a component file because the script tag can't be updated dynamically. For more information, see [Add compiler error if there's a &lt;script&gt; element inside a component (aspnet/Blazor &num;552)](https://go.microsoft.com/fwlink/?linkid=872131).
```
danroth27: I'd rather not reference GitHub issues  as part of our docs. All we really need to say here is that you can't use script tags in components. Instead you need to put them in your *index.html* file. If you want to explain more about why that is, we should pull some of Steve's explanation from the GitHub issue into this doc.

docs/javascript-interop.md
```diff
@@ -127,74 +141,62 @@ Now you can focus inputs in any of your components:
 
 ## Invoke .NET methods from JavaScript functions
 
+### Static .NET method call
+
 To invoke a static .NET method from JavaScript, use the `DotNet.invokeMethod` or `DotNet.invokeMethodAsync` functions. Pass in the identifier of the static method you wish to call, the name of the assembly containing the function, and any arguments. Again, the async version is preferred to support out-of-process scenarios. To be invokable from JavaScript, the .NET method must be public, static, and decorated with `[JSInvokable]`. By default, the method identifier is the method name, but you can specify a different identifier using the `JSInvokableAttribute` constructor. Calling open generic methods isn't currently supported.
```
danroth27: out-of-process -> server-side

docs/javascript-interop.md
```diff
@@ -23,35 +23,49 @@ A Blazor app can invoke JavaScript functions from .NET and .NET methods from Jav
...
+
+.NET methods interop with the JavaScript functions by calling `InvokeAsync<T>` method on `IJSRuntime`.
+
+The sample app uses a pair of C# methods, `Prompt` and `Display`, to invoke the `showPrompt` and `displayWelcome` JavaScript functions:
+
+*JsInteropClasses/ExampleJsInterop.cs*:
+
+[!code-csharp[](./common/samples/2.x/BlazorSample/JsInteropClasses/ExampleJsInterop.cs?name=snippet1&highlight=6-8,14-16)]
 
 The `IJSRuntime` abstraction is asynchronous to allow for out-of-process scenarios. If the app runs in-process and you want to invoke a JavaScript function synchronously, downcast to `IJSInProcessRuntime` and call `Invoke<T>` instead. We recommend that most JavaScript interop libraries use the async APIs to ensure the libraries are available in all Blazor scenarios, client-side or server-side.
```
danroth27: It looks like this text was unchanged, but the PR says it's supposed to also address #257.  I would change out-of-process to server-side and in-process to client-side.

PR #236, opened 2018-09-03
Add PDF Export

### LINKED ISSUE #76: Add PDF Export
Would be nice to be able to export docs to PDF. [DocFx supports PDF export](https://dotnet.github.io/docfx/tutorial/walkthrough/walkthrough_generate_pdf.html), but has a dependency on [wkhtmltopdf](https://wkhtmltopdf.org/) which might be a little hard to integrate with the GitHub pages hosting.

I *think* the way to do this would be to integrated it into the AppVeyor build. There's a Chocolatey package for wkhtmltopdf, and AppVeyor supports Chocolatey installs, so that may be one way to do it. Another is just to use the [wphtmltopdf installer](https://wkhtmltopdf.org/downloads.html).

Once that dependency is in place, we just need to [add a *pdf* section to our docfx.json file](https://dotnet.github.io/docfx/tutorial/walkthrough/walkthrough_generate_pdf.html#step2-add-pdf-section-into-docfxjson) and we'll have a PDF generated.

guardrex:
Fixes #76 

This gets us *close-ish*. The approach here produces:

* *_site_pdf/Blazor.Docs_docs.pdf* - *docs* node only (documentation topics)
* *_site_pdf/Blazor.Docs_api.pdf* - API topics
* *_site_pdf/Blazor.Docs_pdf.pdf* - The *docs* node (documentation topics) with the *Community* page and blog link

The last two are probably the ones we'd want to link to.

Open issues:

* The *includes* (our preview notice) isn't making it thru to the rendered PDFs.
* I didn't link the PDFs yet ... wanted to hear first which ones you want and where you want them linked.


guardrex:
@danroth27 @jongalloway Fairly close to working ...

* I added a pair of links to the PDF files at the top of the docs landing page.
* I opened https://github.com/dotnet/docfx/issues/3446 to ask about the INCLUDES not working.

guardrex:
@danroth27 

I'm not a big fan of 🐌 s....l....o....w 🐌 builds, but I can see how the ability for a reader to take a PDF with them is a nice option.

Do we wait (possibility until the :cow2: come 🏠) for them to [fix the INCLUDES for PDFs](https://github.com/dotnet/docfx/issues/3446)? Our Blazor preview note isn't going to be put into the PDFs.

If you want to proceed, I'll merge this one evening when you're around. I doubt AppVeyor will eat itself, but I wouldn't bet on it. I can revert to :pray: hopefully :pray: repair things if there's a giant :boom:, but that's about all I can do vis-a-vis AppVeyor. It's best if you're on-hand just in case things get really ugly.

danroth27:
Maybe we should just hold off on PDFs until we can actually get the Blazor docs onto docs.microsoft.com.

guardrex:
Silly me ... _we could work around it!_ Instead of using the preview INCLUDE, we just inline that NOTE for each topic. There aren't many topics, so it would be easy to do.

Idk about the build time tho. It's a pain-point waiting for things to build when working on any of these repos, and any browser can save a website for offline viewing.

danroth27:
The more I think about it the more I think we should just wait until we are on docs.microsoft.com. Any objections to just closing this?

guardrex:
@jongalloway This was your :baby:. lol

jongalloway:
I approved; since the work is done I don't see why not ship it, but it's not my call.

guardrex:
@jongalloway ...

Two potential issues:

* It takes a lot longer to build the docs with this in place.
* Unless we physically move the preview notice INCLUDE into the text of the topics, the PDF output won't have the preview notice.

Dan said to :hocho: this for now, and it doesn't sound like ur really *objecting* ... just *suggesting*.

jongalloway:
Right, it was a feature request from me; if @danroth27 wants to cut it, he's in charge. Since you've generated PDFs (in the pull request files), is it possible to preserve them as a snapshot? My usecase was a long plane flight without wifi, slightly old offline docs would have been better than nothing.

guardrex:
> preserve them as a snapshot

Great idea. We still need to address the preview notice. Do I put the notice into each topic and drop the INCLUDE? Seems reasonable ... there aren't that many topics here.

jongalloway:
That makes sense to me. Like you said, there aren't that many topics.

guardrex:
I just found out something interesting about the bug 🐞 ... It's not that it's coming in from an INCLUDE ... it's that it's a note. This means that all of our notes (`[!NOTE]`, `[!WARNING]`, etc.) are busted in the PDFs.

We're not going to be able to overcome this until https://github.com/dotnet/docfx/issues/3446 is fixed. They've marked that up-for-grabs, so I don't think it will be fixed any time soon.

I'm going to go ahead and close.

PR #244, opened 2018-09-14
Add note to remove the global.json file

danroth27:
Lots of folks using 0.5.1 are hitting issues with the *global.json* file included with the project templates. Adding a note to the docs to tell folks to remove the *global.json* file. This issue is fixed in the upcoming 0.6.0 release.

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -45,7 +45,7 @@ Congrats! You just ran your first Blazor app!
 ![Blazor app home page](https://msdnshared.blob.core.windows.net/media/2018/04/blazor-bootstrap-4.png)
 
 > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


guardrex:
@danroth27 Ok to go with an IMPORTANT note there? I put a commit in for it.

danroth27:
@guardrex The issue is specific to not having version 2.1.3xx of the .NET Core SDK installed, so "or later" should be removed. The problem is that the global.json file pins the project to 2.1.3xx, so if you don't have that specific version range then things break. Most folks are hitting this issue because they have a newer version of this SDK installed (like 2.1.4xx) instead of 2.1.3xx. Make sense?

guardrex:
How about we clarify it ... see if the commit works. I rolled your explanation into the passage.

danroth27:
@guardrex Looks good to me

PR #247, opened 2018-09-14
Update instructions to install SDK 2.1.302

danroth27:
Updating getting started instructions to tell folks to specifically install .NET Core SDK 2.1.302 to avoid issues with the global.json files in the templates.


guardrex:
I guess it doesn't matter that the two links go to the same place.

PR #254, opened 2018-09-24
[FIX] <base href="/" />.

srburton:
For working routes to work properly you need to add the base tag in your index.html document, this tag is necessary for the URL to be generated correctly. If you have problems with routes try to add the following example <base href="/" />.

When creating a new PR, please delete this template text and reference the issue number.

To automatically close the issue when the PR is merged, cite the issue number in the first line with ...

Fixes #Issue_Number

To only address an issue with a PR and not automatically close the issue when the PR merges, cite the issue number in the first line with ...

Addresses #Issue_Number

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -37,6 +37,8 @@ Multiple route templates can be applied to a component. In the [sample app](http
 
 [!code-cshtml[](common/samples/2.x/BlazorSample/Pages/BlazorRoute.cshtml?start=1&end=4)]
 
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @srburton

I'm going to make a few tiny changes to the language to conform to our style manual, and I'd like to link this over to our existing coverage for app base path.

guardrex:
I need to update the *App base path* section, too. When the change was made to relative paths, we didn't change the 'set it to an empty string' language over there. I'll set that PR up now.

Thanks again @srburton for flagging this! :rocket:

PR #258, opened 2018-10-01
Update get started instructions for Blazor 0.6.0

### LINKED ISSUE #246: Remove note about removing global.json file once Blazor 0.6.0 ships
Remove this note in the getting started instructions once Blazor 0.6.0 ships:

```
> [!IMPORTANT] 
> The default *global.json* file included in the Blazor project templates may cause the project to fail to load or run if you don't have version 2.1.3xx of the .NET Core SDK installed. The *global.json* file pins the project to 2.1.3xx; so if you don't have that specific version range installed, the project fails to load or run. The workaround is to remove the *global.json* file from the project. 
```

danroth27:
Updates get started instructions for Blazor 0.6.0. Fixes #246

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -17,24 +17,30 @@ uid: client-side/blazor/get-started
 
 ## Setup
 
-1. Install the [.NET Core 2.1 SDK (2.1.302)](https://go.microsoft.com/fwlink/?linkid=873092).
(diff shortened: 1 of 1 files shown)


danroth27:
Hold off on merging this to live until after 0.6.0 is released.

Issue #264, opened 2018-10-05
Update API refs to final Blazor 0.6.0 build


guardrex:
https://github.com/aspnet/Blazor.Docs/tree/master/src

PR #278, opened 2018-10-12
Update api ref to 0.6.0

danroth27:
We're currently missing the XML doc files for Microsoft.JSInterop and Mono.WebAssembly.Interop. We'll add those later once we have them built.


guardrex:
Fixes #264

PR #308, opened 2018-11-09
Concept of Scoped for Server Side Blazor

vertonghenb:
Razor Components or Server Side Blazor, does actually support `Scoped`.


guardrex:
Thanks for putting this in @vertonghenb.

> scoped to the \`Session\`.

I'm not familiar with the details on scoping in Razor Components, but the concept of "session" is entangled with session state here, and I've never heard of the "scoped" lifetime defined in relation to any concept of "session." Shouldn't this read ...

> scoped to the request.

... or is "session" a thing in Razor Components that I just haven't heard about yet?

vertonghenb:
agreed, request is a better term for this.

docs/dependency-injection.md
```diff
@@ -54,7 +54,7 @@ Services can be configured with the following lifetimes:
...
 | [Transient](https://docs.microsoft.com/dotnet/api/microsoft.extensions.dependencyinjection.servicedescriptor.transient) | Whenever a component requires this service, it receives a *new instance* of the service. |
-| [Scoped](https://docs.microsoft.com/dotnet/api/microsoft.extensions.dependencyinjection.servicedescriptor.scoped) | Blazor doesn't currently have the concept of DI scopes. `Scoped` behaves like `Singleton`. Therefore, prefer `Singleton` and avoid `Scoped`. |
+| [Scoped](https://docs.microsoft.com/dotnet/api/microsoft.extensions.dependencyinjection.servicedescriptor.scoped) | Client-side Blazor doesn't currently have the concept of DI scopes. `Scoped` behaves like `Singleton`. Therefore, prefer `Singleton` and avoid `Scoped` in client-side Blazor apps. However, server-side Blazor, also known as Razor Components, supports the `Scoped` lifetime. In a server-side Razor Component, a scoped service registration is scoped to the request. |
```
danroth27: I'm not sure what "to the request" means in this context, since everything in serve-side Blazor is happening over a SignalR connection. @bradygaster do you have a recommendation here?

bradygaster: @danroth27 by "to the request" i presume what's meant is "to the HTTP request" but in the case of Blazor - not sure. Or is your question more about what the scope should be for SignalR hubs being used by server-side Blazor?

guardrex: A comment in the Hubs topic states ...

> Hubs are transient ... Every hub method call is executed on a new hub instance.

... every Blazor SignalR call receives a fresh service instance for any service registered as scoped?

vertonghenb: if I'm not mistaken: from a user point of view, a scoped instance is created for every browser tab a user opens  thus for every SingalR connection. A singleton  on the otherhand is  shared between all tabs for all connected users.

guardrex:
@manuelbl recommended on #354 ...

### Singleton

DI creates a *single instance* of the service. All components requiring this service receive a reference to this instance. Server-side Blazor shares the instance with all user sessions. For client-side Blazor, the instance lives in the client and is therefore session private.

### Scoped 

Server-side Blazor creates a separate service instance for each SignalR connection. It is therefore session private. A page reload creates a new SignalR connection and therefore a new service instance; so it will not live as long as the session. Client-side Blazor doesn't currently have the concept of DI scopes. `Scoped` behaves like `Singleton`, i.e. it is session private.

guardrex:
Cross-ref the DI topic: https://github.com/aspnet/Blazor.Docs/issues/219

docs/dependency-injection.md
danroth27: Yeah, that sounds right. So I think this should say "to the connection" instead of "to the request". Otherwise, looks good to me.

PR #319, opened 2018-11-17
Contributing DocFX Mono updates

guardrex:
Revises the Mono guidelines.

I have a Mac sitting here, but it isn't set up for GH, the repo, DocFX, or Mono. Do I need to test locally (which will take some time), or will the revised guidance ✨ **_Just Work_**:tm: ✨?

Also, this command doesn't quite look right to me (when run from the repo's root) ...

```console
mono $HOME/bin/docfx/docfx.exe serve _site
```

... shouldn't that just be ...

```console
mono $HOME/bin/docfx/docfx.exe --serve
```

... because I thought that `_site` folder is the default if the `--serve` switch is provided.

Thanks @mepda! :rocket:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,7 +108,7 @@ DocFX requires the .NET Framework on Windows or Mono for Linux/macOS.
 
   ```
   alias docfx=mono $HOME/bin/docfx/docfx.exe
-  alias docfx-serve=mono $HOME/bin/docfx/docfx.exe serve _site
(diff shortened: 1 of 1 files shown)


danroth27:
> I have a Mac sitting here, but it isn't set up for GH, the repo, DocFX, or Mono. Do I need to test locally (which will take some time), or will the revised guidance ✨ **_Just Work_**:tm: ✨?

Well, *someone* needs to test these instructions :smiley:. Has anyone tried them out?

guardrex:
Ok ... it works. It just required some quotes around the commands of the aliases.

PR #331, opened 2018-12-02
Update content on use of IJSRuntime

### LINKED ISSUE #326: Update JS interop to document appropriate use of IJSRuntime with server-side Blazor
https://github.com/aspnet/Blazor/issues/1726#issuecomment-441777114

`JSRuntime.Current` always refers to the current user's request. Instead, do the following instead in your component: `@inject IJSRuntime jsRuntime`, and then use jsRuntime to issue the JS interop call instead. That way you'll be sending the info to each client separately.

guardrex:
## ✋ HOLD FURTHER REVIEW 🛑 

**This PR's content has been incorporated into the topic prior to moving it to aspnet/Docs. The topic will be reviewed after the move to aspnet/Docs, and the remarks here will be added to the PR.**

Fixes #326 

* I'm going for a simplification here of what Steve shows in the Material Components sample. As usual, :skull: *death threats* :skull: are welcome when you see the code! :smile: lol
* I assume that `JSRuntime.Current` is still ok for a client-side Blazor app, and I make that utterly clear in these updates (if that's the case; if not, this requires more work).

### DIFF
--- a/docs/javascript-interop.md
+++ b/docs/javascript-interop.md
@@ -76,7 +76,7 @@ The following Blazor component:
     // David Krumholtz on IMDB: https://www.imdb.com/name/nm0472710/
 
     private MarkupString ConvertedText =
-        new MarkupString("Select the <b>Convert Text</b> button.");
+        new MarkupString("Select the <b>Convert Array</b> button.");
 
     private uint[] QuoteArray = new uint[]
         {
--- a/docs/javascript-interop.md
+++ b/docs/javascript-interop.md
@@ -30,7 +30,7 @@ For server-side Blazor apps:
 * Multiple user requests are processed by the server-side app. Don't call `JSRuntime.Current` in a component to invoke JavaScript functions.
 * Inject the `IJSRuntime` abstraction and use the injected object to issue JavaScript interop calls.
 
-The following example is based on [TextDecoder](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder), an experimental JavaScript-based decoder. The example demonstrates how to invoke a JavaScript function from a C# method. The JavaScript function accepts a byte array from a C# method, decodes the array, and returns the text to the Blazor component for display.
+The following example is based on [TextDecoder](https://developer.mozilla.org/docs/Web/API/TextDecoder), an experimental JavaScript-based decoder. The example demonstrates how to invoke a JavaScript function from a C# method. The JavaScript function accepts a byte array from a C# method, decodes the array, and returns the text to the Blazor component for display.
 
 Inside the `<head>` element of *wwwroot/index.html*, provide a function that uses `TextDecoder` to decode a passed array:
 
--- a/docs/javascript-interop.md
+++ b/docs/javascript-interop.md
@@ -43,7 +43,7 @@ Inside the `<head>` element of *wwwroot/index.html*, compose a function that:
     div: {
       init: function (elem, content) {
         elem.innerHTML = content;
-      }
+    }
   }
 };
 </script>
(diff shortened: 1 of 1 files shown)


guardrex:
@SteveSandersonMS This is another one that needs an :eye: ... perhaps even more important, since it's a doc bug fix.

docs/javascript-interop.md
```diff
@@ -23,9 +23,68 @@ A Blazor app can invoke JavaScript functions from .NET and .NET methods from Jav
...
+
+The following example component in the server-side Blazor app:
+
+* References a DIV element using the `ref` attribute in the component's markup.
+* Invokes the JavaScript function using `jsRuntime` with the referenced DIV element and the markup content assigned to the `DivContent` property.
+
+```cshtml
+@page "/"
+@using Microsoft.JSInterop;
+@inject IJSRuntime jsRuntime;
```
danroth27: Injected properties should be camel cased.

docs/javascript-interop.md
```diff
@@ -23,9 +23,68 @@ A Blazor app can invoke JavaScript functions from .NET and .NET methods from Jav
...
+
+    protected override Task OnAfterRenderAsync()
+        => jsRuntime.InvokeAsync<object>(
+            "Index.div.init",
+            divElem,
+            DivContent);
+}
+```
+
+For client-side Blazor apps, the `IJSRuntime` abstraction is accessible from `JSRuntime.Current`, which refers to the current user's request. Because there's only a single user of a client-side Blazor app, using `JSRuntime.Current` to invoke a JavaScript function works normally. Only use `JSRuntime.Current` in client-side Blazor apps.
```
danroth27: @SteveSandersonMS What are your thoughts on the guidance we should give for when to use `JSRuntime.Current`? Should it ever be used?

docs/javascript-interop.md
```diff
@@ -23,9 +23,68 @@ A Blazor app can invoke JavaScript functions from .NET and .NET methods from Jav
...
+
+* Receives a reference to the element and the markup content.
+* Applies the content to the element by assigning to the element's `innerHTML` property.
+
+```html
+<script>
+  window.Index = {
+    div: {
+      init: function (elem, content) {
+        elem.innerHTML = content;
```
danroth27: I'd prefer to not show using JS interop to manipulate the DOM. In Blazor apps you typically shouldn't be messing with the DOM directly. Can we show calling some sort of browser API API instead?

docs/javascript-interop.md
danroth27: Oh, yeah, that 😄.

guardrex:
@danroth27 I swapped that first example for a JS function that uses `XMLHttpRequest`. Don't like it tho. It doesn't pass anything and doesn't return anything to speak of (just a bool that the FN was called). I'll look at some other options.

danroth27:
> I swapped that first example for a JS function that uses `XMLHttpRequest`.

Yeah, that's not ideal either, because we already provide you an `HttpClient` that does the JS interop under the covers to call the browser `fetch` API. How about calling `prompt`?

guardrex:
I had an idea for a FN that uses a JS call to change some text encoding, but it :boom: on me in a test.

We have the `prompt` example in the expanded example that's part of the sample app just below this. We could just cut drop this first example and go straight to the expanded examples.

The idea behind this first example was to show something really simple and quick.

guardrex:
I think I'll try that encoding thingy one more time. Maybe I can get it to work. If not tho, we can drop this example.

guardrex:
Ah, ha! I guess I just wasn't calling `StateHasChanged();` before. 🤦‍♂️

I'll have a better example up shortly. This one uses `TextDecoder`, passes something in, and receives a string back for display ... and it's super-simple to grok.

guardrex:
@danroth27 kk ... I think that's better. I might need to add a warning about `TextDecoder` support. 🤔 ... in development for Edge tho ... https://caniuse.com/#search=TextDecoder

PR #333, opened 2018-12-06
Blazor templates have to be installed first

retifrav:
By default `dotnet new blazor` will fail due to the missing templates:

```
No templates matched the input template name: blazor.
```

...so those have to be installed first.


danroth27:
This is covered in the [Get Started](https://blazor.net/docs/get-started.html) steps, but I think it's fine to repeat here.

PR #342, opened 2018-12-16
Add https://www.highcharts.com/

Hantse:
Adding library for Highcharts


guardrex:
@Hantse We haven't heard back, so I'm going to close. If you make the requested changes, ping me back here to re-open.

PR #343, opened 2018-12-17
Component libraries

### LINKED ISSUE #70: Add topic on building component libraries
We should have a topic around the new component library template that shows how to build, package, and share libraries of components.

See https://blogs.msdn.microsoft.com/webdev/2018/04/17/blazor-0-2-0-release-now-available/ for some existing content.

stimms:
Adding top-level section on using component libraries. Includes

How to create a library
How to add it to solution
How to consume it
How to pack it for nuget
First docs commit so I'm very open to suggestions.

Fixes #70

### DIFF
--- a/docs/components/component-libraries.md
+++ b/docs/components/component-libraries.md
@@ -5,7 +5,7 @@ description: Discover how Blazor components can be included in a Blazor app from
 manager: wpickett
 ms.author: riande
 ms.custom: mvc
-ms.date: 12/14/2018
+ms.date: 12/17/2018
 ms.prod: asp.net-core
 ms.technology: aspnet
 ms.topic: article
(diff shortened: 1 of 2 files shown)


guardrex:
Thanks @stimms ... I'll take look later today.

guardrex:
@stimms I did a UE+guidelines pass. Make sure that I didn't break the meaning of anything. The only spot I'm concerned about is ...

> The `@addTagHelper` directive can be included in *_ViewImport.cshtml* to make the components available for an entire project or scoped to a single page or set of pages within a folder.

You're saying that "scoped to a single page or set of pages" applies when the *_ViewImport.cshtml* file is within the folder where the pages reside, correct?

PR #354, opened 2018-12-26
Improved description of service scope and lifetime

manuelbl:
Service scope and lifetime is different for client and server side Blazor. The updated description goes into more detail


guardrex:
Hello @manuelbl ... This is a duplicate of #308. Engineers will probably come back in the new year and pick up with the discussion on what to do with the text. I'll move your proposed changes to that issue for further discussion and close this one. Track on that PR ... we'll end up only taking one set of changes for whatever the update is going to be.

PR #355, opened 2018-12-26
Describe default service for server-side Blazor

manuelbl:
`HttpClient` is not provided for server-side Blazor.

### DIFF
--- a/docs/dependency-injection.md
+++ b/docs/dependency-injection.md
@@ -60,12 +60,14 @@ Blazor's DI system is based on the DI system in ASP.NET Core. For more informati
 
 ## Default services
 
-Blazor provides default services that are automatically added to the service collection of an app. The following table shows a list of the default services currently provided by Blazor's [BrowserServiceProvider](/api/Microsoft.AspNetCore.Blazor.Browser.Services.BrowserServiceProvider.html).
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @manuelbl.

We probably won't need to expand the list (but we could). For reference ...

* https://github.com/aspnet/AspNetCore/blob/62dd8f4bd999e4181cb7e5d9fd8901eee943af30/src/Components/src/Microsoft.AspNetCore.Components.Server/DependencyInjection/ServerSideBlazorServiceCollectionExtensions.cs#L111-L130
* https://github.com/aspnet/AspNetCore/blob/ceb1c8e562562b2655622b571d2a9ad083806ad5/src/Components/blazor/src/Microsoft.AspNetCore.Blazor/Hosting/WebAssemblyHostBuilder.cs#L86-L111

If the list isn't expanded, we probably should do one of the following:

* Add one or two columns to the table to indicate which services go with which hosting model.
* Add a line to each of the descriptions to indicate which services go with which hosting model. *I'm leaning towards this approach.*

If the list is expanded, a better layout for this content might be to have a subsection for each hosting model.

Also note that `BrowserServiceProvider` is going away. I think we should drop the name now to harden this section up for 0.8.0.

I'll make an edit to see how it looks.

guardrex:
@rstropek Have a minute to put an :eye: on this?

PR #367, opened 2019-01-03
Template name fix

treefishuk:
Incredibly minor but the template dialog in Visual Studio says 'in' rather than 'on' ASP.NET Core


guardrex:
lol ... I think their choice is grammatically incorrect. Apps run **_ON_** frameworks ... not *in* them. :trollface: lol

PR #376, opened 2019-01-15
Data binding methods with parameters. Fixes #296

### LINKED ISSUE #296: Data binding methods with parameters
I would like to make an improvement to the Data binding section of Components. It currently does not indicate how to pass parameters in methods. This can be accomplished using the `@(() => Method(parameter))` syntax. This could also be supplemented by an example.

MorneZaayman:


MorneZaayman:
Saw comment regarding  `@(async (e) => Method(e, parameter))` on the issue. Will reopen once I've added this.

PR #379, opened 2019-01-18
Data binding methods with parameters 

### LINKED ISSUE #296: Data binding methods with parameters

MorneZaayman:
Fixes #296

I added the updates at the bottom of **Data Binding** section as suggested in the issue, but after writing it I think it is more suited to go under the **Event Handling** section.

### DIFF
--- a/docs/components/index.md
+++ b/docs/components/index.md
@@ -5,7 +5,7 @@ description: Learn how to create and use Blazor components, the fundamental buil
 manager: wpickett
 ms.author: riande
 ms.custom: mvc
-ms.date: 01/26/2019
+ms.date: 01/29/2019
 ms.prod: asp.net-core
 ms.technology: aspnet
 ms.topic: article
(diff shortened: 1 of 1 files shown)


Andrzej-W:
In ***Multiple parameters can also be passed to methods*** you have an example from previous section with only one parameter.

guardrex:
Thanks @MorneZaayman ... I'll take a look this weekend and get back to you.

MorneZaayman:
@Andrzej-W Thanks, looks like I pasted the one example twice instead of the correct example the second time. I've fixed that up.

guardrex:
**_I haven't forgotten about this!_** 🏃😅 I plan to look on Saturday &ndash; if my head doesn't explode :boy:💥:smile:.

MorneZaayman:
Haha , no problem. When I ever you get a chance 👍🏻

Regards
Morné Zaayman

________________________________

guardrex:
@MorneZaayman ... Check my updates on the last commit and make sure that I didn't break anything.

MorneZaayman:
@guardrex Looks good to me.

Andrzej-W:
@guardrex, @MorneZaayman it looks that new content is in wrong place. I fully agree with this initial comment:
> I added the updates at the bottom of Data Binding section as suggested in the issue, but after writing it I think it is more suited to go under the Event Handling section.

At the bottom of `Event Handling` section we currently have
> Lambda expressions can also be used:
`<button onclick="@(e => Console.WriteLine("Hello, world!"))">Say hello</button>`

I think we can remove these 2 lines and move new content here, but without **Data binding methods with parameters** subtitle. Event handling and data binding are two different things.

guardrex:
@Andrzej-W Are you only referring to the 3rd example in the new content, or to all of this? Perhaps, the first examples should change to something more data binding **_-ish_**, while the 3rd example should be moved down to the event handling section. If the thrust of the new content is **_methods with parameters_**, I'm asking if this is useful in *both* scenarios, thus asking if it should be covered in both sections?

docs/components/index.md
```diff
@@ -190,6 +190,81 @@ If the value of the `ParentYear` property is changed by selecting the button in
 <p>Year: 1986</p>
 ```
 
+**Data binding methods with parameters**
```
danroth27: This isn't really about data binding. This is all event handling. Also, the event handler don't take additional parameters. But the code *inside* the event handler can contain whatever you like including code that closes over values in scope

We already mention at the end of the event handling section that you can use lambda syntax. I think it would be sufficient to expand on that statement to say that it's often convenient to close over additional value (like the current item in a `foreach` loop).

guardrex:
@danroth27 Something like that? (see last commit)

PR #401, opened 2019-02-01
Documentation for Fallback Components

TheFanatr:
This change adds documentation about the `FallbackComponent` parameter now present on the `Router` component. It enables support for rendering a specific component when none are resolved for a certain route, and is now usable as of ASP.NET Core Preview 2; as such, I feel that it would be sensible to add documentation for it's usage.


guardrex:
Thanks @TheFanatr ... We're in the process of moving the docs to the official repo. I'll move this PR over shortly and merge it from there. I'll ping you on the PR.

**Leave this open so that I don't lose track of it.**

